PR #42, opened 2017-02-21
Reorganize wrenhandler to consistently handle errors

ericmjonas:
This ended up being a bit bigger of a change than I expected, but now when parts of wrenhandler raise exceptions, we catch them and serialize a bit of information about them and return it to the client. Note this is NOT the same thing as the _python function_ that we execute raising an exception, those have always worked. 

This is for errors like:
- Job takes too long and we timeout and return an error to the client
- pywren version mismatch
- s3 read/write errors

Both timeout errors and version mismatch errors are cleanly handled by the client library; other errors have the string representation of their exception passed through. This isn't the best from a UX perspective, but we can always wrap them cleanly, and at least the job doesn't just hang/fail. 

For testing purposes, map now lets you `overwrite_invoke_args` to break parts of the invocation (like the version string) during testing. This theoretically will let us expand testing coverage of the handler in the future. 

As an added benefit this PR introduces version checking, and we take versioning a little bit more seriously now.


pywren/wren.py
```diff
@@ -164,6 +166,11 @@ def invoke_with_keys(self, s3_func_key, s3_data_key, s3_output_key,
         logger.info("call_async {} {} lambda invoke ".format(callset_id, call_id))
         lambda_invoke_time_start = time.time()
 
+        # overwrite args
+        # this is entirely for debugging various wrenhandler errors
```
shivaram: This is only used for testing - Is my reading right ? If so you can put that in the comment

pywren/wren.py
```diff
@@ -515,7 +523,17 @@ def result(self, timeout=None, check_only=False, throw_except=True):
         self._invoke_metadata['status_done_timestamp'] = time.time()
         self._invoke_metadata['status_query_count'] = self.status_query_count
             
-        # FIXME check if it actually worked all the way through 
+        if call_status['exception'] is not None:
+            # the wrenhandler had an exception
+            exception_str = call_status['exception']
+            print(call_status.keys())
+            exception_args = call_status['exception_args']
+            if exception_args[0] == "WRONGVERSION":
```
shivaram: This is a minor nit - but it might good to put these strings `WRONGVERSION` and `OUTATIME` in a separate class (something like a enum ?)

pywren/wren.py
```diff
@@ -540,7 +558,16 @@ def result(self, timeout=None, check_only=False, throw_except=True):
             self._return_val = call_invoker_result['result']
             self._state = JobState.success
         else:
-            if call_status['process_timeout_killed']:
+            if call_status['exception'] is not None:
+                # the wrenhandler had an exception
+                exception_str = call_status['exception']
+                print(call_status.keys())
+                exception_args = call_status['exception_args']
+                if exception_args[0] == "VERSIONERROR":
```
shivaram: This should be `WRONGVERSION` ? (My previous comment was prescient !)

pywren/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.10rc0"
```
shivaram: I think `0.1rc0` makes more sense ? `0.10` might be the 10th release etc.

pywren/wren.py
ericmjonas: Ha, that was, if only because this code path is entirely dead! I removed it

pywren/wren.py
ericmjonas: I think this is a good suggestion, although a bit overkill for the two errors we have right now -- I think ultimately the right thing is going to be defining our own exception classes, but I'm tabling that right now -- we have tests that cover all these edge cases, which is the main thing I care about.

PR #43, opened 2017-02-21
Validate runtime python version

### LINKED ISSUE #40: The runtime situation is a mess
There's no programmatic way of knowing which runtime you're using
There's no good way of making sure you don't accidentally try to run a python2 job with the python3 runtime
It's easy to accidentally deploy a broken runtime.

### LINKED ISSUE #41: Create package versioning and check with runtime
We would like to make sure that the lambda `wrenhandler.py` and the locally-installed library always are the same version. 

Create a `__version__` for the package, and have the lambda runtime compare its version with that.

ericmjonas:
Closes #41 and effectively closes #40 for this milestone.


ooq:
LGTM.

A quick yet nice improvement is to pick the right runtime during create_config?
I can send a PR if that's reasonable.

ericmjonas:
@ooq Yes that would be awesome!

PR #47, opened 2017-02-22
Handle un-pickleable exceptions

### LINKED ISSUE #18: Fix directories so that the conda command-line utils have right path
- [ ] Are the conda utils / is the anaconda environment in the runtime? 
- [ ] Can we conda install or pip install something?

ericmjonas:
While attempting to resolve #18 I discovered that subprocess would raise an exception when invoked with a non-existant command, but that things would then fail when unpickling. Closer investigation showed that some custom python exceptions don't invoke the base class constructor and as a result they can't be unpickled. Now, pickling exceptions is a weird thing, so I guess I understand why this happens. 

To resolve, when an exception is triggered, `jobrunner.py` tries to save it and then reload it. If this fails we then save string versions of the exception and do some cleanup on the client side. It's not ideal, but an exception still propagates, which is what matters. 

Note this PR is a branch from  `release-setup` and should only be merged after.

### DIFF
--- a/pywren/jobrunner.py
+++ b/pywren/jobrunner.py
@@ -38,11 +38,34 @@ try:
 except Exception as e:
     exc_type, exc_value, exc_traceback = sys.exc_info()
     traceback.print_tb(exc_traceback)
-    pickle.dump({'result' : e, 
-                 'exc_type' : exc_type, 
-                 'exc_value' : exc_value, 
-                 'exc_traceback' : exc_traceback, 
-                 'sys.path' : sys.path, 
-                 'success' : False}, 
-                open(out_filename, 'wb'), -1)
+
+    # Shockingly often, modules like subprocess don't properly 
+    # call the base Exception.__init__, which results in them 
+    # being unpickleable. As a result, we actually wrap this in a try/catch block
+    # and more-carefully handle the exceptions if any part of this save / test-reload
+    # fails
+
+    try:
+        pickle.dump({'result' : e, 
+                     'exc_type' : exc_type, 
+                     'exc_value' : exc_value, 
+                     'exc_traceback' : exc_traceback, 
+                     'sys.path' : sys.path, 
+                     'success' : False}, 
+                    open(out_filename, 'wb'), -1)
+        # this is just to make sure they can be unpickled
+        pickle.load(open(out_filename, 'rb'))
+
+    except Exception as pickle_exception:
+        pickle.dump({'result' : str(e), 
+                     'exc_type' : str(exc_type), 
+                     'exc_value' : str(exc_value), 
+                     'exc_traceback' : exc_traceback, 
+                     'exc_traceback_str' : str(exc_traceback), 
+                     'sys.path' : sys.path, 
+                     'pickle_fail' : True, 
+                     'pickle_exception' : pickle_exception, 
+                     'success' : False}, 
+                    open(out_filename, 'wb'), -1)
+        
     
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -539,6 +539,7 @@ class ResponseFuture(object):
                 raise Exception(exception_str, *exception_args)
         
         call_output_time = time.time()
+        print("calling get_call_output")
         call_invoker_result = get_call_output(self.callset_id, self.call_id, 
                                               AWS_S3_BUCKET = self.s3_bucket, 
                                               AWS_S3_PREFIX = self.s3_prefix,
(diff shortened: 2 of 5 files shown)


pywren/jobrunner.py
```diff
@@ -38,11 +38,34 @@ def b64str_to_bytes(str_data):
...
+                    open(out_filename, 'wb'), -1)
+        # this is just to make sure they can be unpickled
+        pickle.load(open(out_filename, 'rb'))
+
+    except Exception as pickle_exception:
+        pickle.dump({'result' : str(e), 
+                     'exc_type' : str(exc_type), 
+                     'exc_value' : str(exc_value), 
+                     'exc_traceback' : exc_traceback, 
+                     'exc_traceback_str' : str(exc_traceback),
```
shivaram: do we need `exc_traceback` and `exec_traceback_str` ? also does `exc_traceback` always get pickled correctly ?

pywren/wren.py
```diff
@@ -539,6 +539,7 @@ def result(self, timeout=None, check_only=False, throw_except=True):
                 raise Exception(exception_str, *exception_args)
         
         call_output_time = time.time()
+        print("calling get_call_output")
```
shivaram: is this useful to print each time or just something useful for debugging ?

pywren/wren.py
```diff
@@ -556,28 +557,36 @@ def result(self, timeout=None, check_only=False, throw_except=True):
...
+
+
         if call_success:
 
             self._return_val = call_invoker_result['result']
             self._state = JobState.success
-        else:
+            return self._return_val
+
+        elif call_success == False and throw_except:
```
shivaram: `call_success` is always False here - so this can be `if throw_except` (since you return from previous if)

pywren/wren.py
```diff
@@ -556,28 +557,36 @@ def result(self, timeout=None, check_only=False, throw_except=True):
...
+
             self._exception = call_invoker_result['result']
             self._traceback = (call_invoker_result['exc_type'], 
                                call_invoker_result['exc_value'], 
                                call_invoker_result['exc_traceback'])
 
             self._state = JobState.error
+            if call_invoker_result.get('pickle_fail', False):
+                logging.warning("there was an error pickling the resulting exception: {}".format(call_invoker_result['exc_value']))
```
shivaram: minor: but it might also be worth while printing the pickle_exception here

pywren/jobrunner.py
```diff
@@ -38,11 +38,34 @@ def b64str_to_bytes(str_data):
...
+    try:
+        pickle.dump({'result' : e, 
+                     'exc_type' : exc_type, 
+                     'exc_value' : exc_value, 
+                     'exc_traceback' : exc_traceback, 
+                     'sys.path' : sys.path, 
+                     'success' : False}, 
+                    open(out_filename, 'wb'), -1)
+        # this is just to make sure they can be unpickled
+        pickle.load(open(out_filename, 'rb'))
```
ooq: should make sure fp for `open(out_filename, 'wb')` is closed before the load?

pywren/jobrunner.py
ericmjonas: `exc_traceback` seems to be pickled on the server side much more reliably, but may still fail on the unpickling -- thus wanting a string copy of it that we can use no matter what.

pywren/wren.py
ericmjonas: thanks, debugging detritus

pywren/jobrunner.py
shivaram: Then it'd be safer to not include the traceback and just keep the traceback str ? I just worry we will have cascading failures

ooq:
Is this issue resolved? @ericmjonas 
I'm asking because I'm running into this problem. I guess to make things simple, we can just stringfy all of these objects that have potential pickle problems? The only benefit of keeping them as objects seem to be that we can reraise the exception later?

ericmjonas:
Ha, it was supposed to be. I think this is always going to be a challenge. I had thought we had chased out all the edge cases here, but it's hard to test. 

I really like the user experience of being able to reraise the exception and give a sane traceback, and have found it quite useful. For stability this might be something we ultimately have to abandon, but I hope not. 

Is there any chance you have some captured output /  logs from when this happens, or better yet a reproducible test case? :)

ericmjonas:
Also Github won't let me reopen the issue because it was on a branch? I'm going to create a new issue for us to keep tracking this.

Issue #50, opened 2017-02-22
create_config script should pick the correct runtime based on python version


ooq:
right now it defaults to 2.7, which could make it difficult for python 3 users to try out pywren

PR #52, opened 2017-02-22
[Issue #50] create_config selects runtime based on local python version.

ooq:
Addressing #50. Replacing https://github.com/ericmjonas/pywren/pull/51 due to testing issue.

Testing for this PR (and the cli script in general) should be added as separate PR in the future.

@ericmjonas @shivaram


ericmjonas:
+1 assuming it passes tests :) 

On line 46 of `.travis.yaml` we explicitly set pythonver, I think this means we can stop doing that? 

Once that's done I'll happily merge.

ooq:
@ericmjonas Fixed.

Issue #53, opened 2017-02-22
Encode and serialize python modules as UTF-8


ericmjonas:
UTF8 is a perfectly valid format for python files, and JSON should also handle it quite fine, too. Make sure we are doing this explicitly.

PR #57, opened 2017-02-22
Function serialize cleanup

### LINKED ISSUE #48: s3util func key calls it lambda.pickle
1. It's now JSON encoded
2. It should be called func not lambda.

ericmjonas:
This is a bit of a big one, around job serialization issues that we found earlier: 
1. Switch to making sure we properly utf-8 encode the python modules that we're sending to lambda. Added some tests to make sure we don't break this in the fuutre. 
2. Replaced some hard-coded paths, some old key names, etc. 
3. Updated the list of modules included in the runtime. 

closes #48, #49 , #53

### DIFF
--- a/pywren/cloudpickle/preinstalls.py
+++ b/pywren/cloudpickle/preinstalls.py
@@ -316,5 +316,6 @@ modules = [
 ('scipy', False), 
 ('sklearn', False), 
 ('cvxpy', False), 
-('cvxopt', False)
+('cvxopt', False),
+('future', False)
 ]
--- a/pywren/cloudpickle/preinstalls.py
+++ b/pywren/cloudpickle/preinstalls.py
@@ -317,5 +317,10 @@ modules = [
 ('sklearn', False), 
 ('cvxpy', False), 
 ('cvxopt', False),
-('future', False)
+('future', False), 
+('six', False), 
+('glob2', False), 
+('tblib', False),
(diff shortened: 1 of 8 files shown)


tests/extmoduleutf8.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+
+TEST_STR = "ᚠᛇᚻ᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢᚱ᛫ᚠᛁᚱᚪ᛫ᚷᛖᚻᚹ"
+def unicode_str(x):
+    return TEST_STR
+
+
```
shivaram: lots of empty lines :)

ericmjonas: I think the right thing to do is to start using PEP-8

pywren/cloudpickle/preinstalls.py
```diff
@@ -316,5 +316,11 @@
 ('scipy', False), 
 ('sklearn', False), 
 ('cvxpy', False), 
-('cvxopt', False)
+('cvxopt', False),
```
ericmjonas: Yeah, I think that's a good plan, although I think in the very near future this will be replaced / obsolete by the introspection stuff

PR #65, opened 2017-02-26
Support wait ANY_COMPLETED

ooq:
Support `ANY_COMPLETED` option in `wait()`.
Testing for this feature is a bit tricky: due to the variance of network/lambda performance, the test could fail (although I believe the current setup should pass >95% of the time). Therefore, it will be ignored in CI (using a `skip` marker).
In the future, we should have dummy executor for this test.
@ericmjonas @shivaram


ooq:
@ericmjonas Nice trick! Added.

ooq:
Also I'm enabling those tests.

PR #72, opened 2017-02-28
Runtime Sharding

### LINKED ISSUE #71: `result(throw_except=False)` can still throw out-of-time exceptions
We need to fix this handling of exceptions, so that result can be used to poll status for jobs that will fail

ericmjonas:
There's a lot in this . we:
1. Fixed some exception handling errors (closes #71)
2. added extra timing metrics so we can track runtime download, func download, and data download times
3. Implemented runtime sharding, where we randomly pick one of N runtime s3 keys to invoke. It's transparent to the user, and off by default.

### DIFF
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -46,7 +46,7 @@ def default_executor(**kwargs):
         return dummy_executor(**kwargs)
     return lambda_executor(**kwargs)
         
-def lambda_executor(config= None, job_max_runtime=280):
+def lambda_executor(config= None, job_max_runtime=280, shard_runtime=False):
 
     if config is None:
         config = wrenconfig.default()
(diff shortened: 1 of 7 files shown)


ooq:
line 503 in wren.py:
`         if self._state == JobState.error:
             raise self._exception
`
Also check `throw_except` here?

ooq:
I agree with what you put in issue#75 that auto-sharding should be on by default. Maybe this should be a feature that comes with specific runtimes? As there is additional work per runtime to support sharding (you have to create copies, upload them, etc). The metadata for sharding support of a runtime could be put in .meta.json.

ooq:
Ideally we want 3 separate PRs for this...

PR #73, opened 2017-02-28
Prototype rate limit executor in issue#66

ooq:
Current changes to `map()` interface is not ideal, but hopefully this gives us a starting point.
@ericmjonas @shivaram

### DIFF
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -211,7 +211,8 @@ class Executor(object):
 
     def map(self, func, iterdata, extra_env = None, extra_meta = None, 
             invoke_pool_threads=64, data_all_as_one=True, 
-            use_cached_runtime=True, overwrite_invoke_args = None):
+            use_cached_runtime=True, overwrite_invoke_args = None,
+            callset_id = s3util.create_callset_id(), custom_ids = []):
         """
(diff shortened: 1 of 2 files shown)


ooq:
I'm not so sure about the build failure here...

pywren/wren.py
```diff
@@ -211,7 +211,8 @@ def agg_data(self, data_strs):
 
     def map(self, func, iterdata, extra_env = None, extra_meta = None, 
             invoke_pool_threads=64, data_all_as_one=True, 
-            use_cached_runtime=True, overwrite_invoke_args = None):
+            use_cached_runtime=True, overwrite_invoke_args = None,
+            callset_id = s3util.create_callset_id(), custom_ids = []):
```
ericmjonas: Python will only ever call create_callset_id once here, upon module creation.

ericmjonas:
Now that I'm looking closer, this represents actually a petty substantial change in the way that pywren works behind the hood -- can we have a conversation about this and the right way to do it?

ooq:
@ericmjonas Yes, I agree. I also feel wrapping around executors might not be easy/clean. Let's discuss before proceeding more.

Issue #84, opened 2017-03-09
Modules do not serialize in ipython notebook


Vaishaal:
The following code works as a script but errors in a jupyter notebook

```
import pywren
import dummy_module
pwex = pywren.default_executor()
results = pwex.map(dummy_module.test, [0])
print(results[0].result())
```
Script output:

> 0

> 

Notebook output:

> ---------------------------------------------------------------------------
> ImportError                               Traceback (most recent call last)
> <ipython-input-23-6212b841d0c1> in <module>()
>       3 pwex = pywren.default_executor()
>       4 results = pwex.map(dummy_module.test, [0])
> ----> 5 print(results[0].result())
> 
> /data/vaishaal/anaconda2/lib/python2.7/site-packages/pywren/wren.pyc in result(self, timeout, check_only, throw_except)
>     597             else:
>     598                 # reraise the exception
> --> 599                 reraise(*self._traceback)
>     600         else:
>     601             return None  # nothing, don't raise, no value
> 
> /var/task/jobrunner.py in <module>()
> 
> ImportError: No module named dummy_module
> 

Contents of dummy_module.py:
```
def test(x):
    return 0
```

Vaishaal:
Fixed by #85

PR #85, opened 2017-03-09
fix to issue 84, modules work in interactive mode

Vaishaal:
Solves issue #84 

Relative paths were not handled properly when serializing modules.

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -103,8 +103,12 @@ class SerializeTest(unittest.TestCase):
             print(f[0])
 
 class InteractiveTest(unittest.TestCase):
-    def setUp(self):
-        self.wrenexec = pywren.default_executor()
+
+    ''' pywren handles module serialization slightly differently in interactive mode vs regular script mode
+        this is important because prior to f7a900f0a3aa185406abce2af3a1697759f464da module imports were failing 
+        in interactive mode, thus the common usecase of launching pywren in an jupyter notebook would fail.
+        This is a (somewhat brittle) test that simulates an interactive python session using the -c flag
+    '''
 
     def test_simple(self):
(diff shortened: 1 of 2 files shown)


tests/test_modules.py
```diff
@@ -101,3 +102,15 @@ def func(x):
         for f in info['pkg_ver_list']:
             print(f[0])
 
+class InteractiveTest(unittest.TestCase):
+    def setUp(self):
```
ericmjonas: This wren executor isn't used, can we get rid of it?

tests/test_modules.py
```diff
@@ -101,3 +102,15 @@ def func(x):
         for f in info['pkg_ver_list']:
             print(f[0])
 
+class InteractiveTest(unittest.TestCase):
```
ericmjonas: Can we add a little bit of documentation here on what we're trying to test, and explain the difference between interactive vs regular mode?

tests/test_modules.py
```diff
@@ -101,3 +102,15 @@ def func(x):
...
+    def setUp(self):
+        self.wrenexec = pywren.default_executor()
+
+    def test_simple(self):
+        ret = os.system("python -c \"import pywren;import extmodule;pwex = pywren.default_executor();results = pwex.map(extmodule.foo_add, [0]);print(results[0].result())\"")
```
ericmjonas: According to the python docs: "The subprocess module provides more powerful facilities for spawning new processes and retrieving their results; using that module is preferable to using this function. See the Replacing Older Functions with the subprocess Module section in the subprocess documentation for some helpful recipes."

Can we try using `subprocess` instead?

tests/test_modules.py
Vaishaal: fixed.

Issue #86, opened 2017-03-09
Standardize configuration for executors


shivaram:
As a part of this also investigate if we can make the user-specified configuration more consistent with where it is used

PR #88, opened 2017-03-10
Move runtime sharding into config variable

### LINKED ISSUE #75: Turn on runtime sharding by default
1. This should probably be on by default

2. It's not clear if it should really be an argument to the executor or in the config or what.

shivaram:
This closes #75

### DIFF
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -6,6 +6,9 @@ except:
     import pickle
 from tblib import pickling_support
 import logging
+import botocore
+import glob2
+import random
 import os
 pickling_support.install()
 
@@ -70,6 +73,7 @@ def remote_executor(config= None, job_max_runtime=3600):
                     job_max_runtime)
 
 standalone_executor = remote_executor
+
 
 def get_all_results(fs):
     """
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -71,7 +71,6 @@ def remote_executor(config= None, job_max_runtime=3600):
 
 standalone_executor = remote_executor
 
-
 def get_all_results(fs):
     """
     Take in a list of futures and block until they are repeated, 
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -32,7 +32,7 @@ def default_executor(**kwargs):
     return lambda_executor(**kwargs)
 
 
-def lambda_executor(config= None, job_max_runtime=280):
+def lambda_executor(config=None, job_max_runtime=280):
 
     if config is None:
         config = wrenconfig.default()
--- a/pywren/wren.py
+++ b/pywren/wren.py
@@ -30,9 +30,8 @@ def default_executor(**kwargs):
     elif executor_str == 'dummy':
         return dummy_executor(**kwargs)
     return lambda_executor(**kwargs)
-
-
-def lambda_executor(config= None, job_max_runtime=280, shard_runtime=False):
+        
+def lambda_executor(config= None, job_max_runtime=280):
 
     if config is None:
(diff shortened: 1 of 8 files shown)


shivaram:
Couple of things that came to my mind while making this change

- While we generate the default config, should we set `num_shards` to 50 ?
- We need to document what all the valid config variables are and where they are used. I will do this as a part of #86

ooq:
I actually meant that `num_shards` should completely be in `[runtime_name].meta.json` that is stored in S3 and the client code should be completely hidden from sharding. 
This would also make switching between runtimes easier.

shivaram:
Ah I see. Ok - @ericmjonas Is that what you had in mind as well ? I can make that change but I think we need to be careful then to make sure runtimes generated before pywren 0.2 are still compatible etc.

ericmjonas:
Yes that's what i was thinking -- do we care much about previous-generation
runtimes? can we default try and get the key in the json and fallback
gracefully to nonsharding?

shivaram:
Yep - I will try to do that. It'll require a change to the runtimes repo, but it should be doable.

shivaram:
@ericmjonas @ooq I updated this, but this depends on https://github.com/pywren/runtimes/pull/2 -- So I'd suggest reviewing that first. 

This is backwards compatible in that if the metadata doesn't have a `urls` field we just fall back to the base runtime url. The one catch though is that all the lambdas will need to access the metadata. I considered shipping that as a part of the job description, but I was worried if it would be too big for the 128KB limit ?

ericmjonas:
@shivaram I'm a little confused here why the lambdas need access to the metadata, can't we just pass them a runtime key as part of the invoker data?

ericmjonas:
If we have the executor, upon creation, read the `meta.json` from s3, then we will: 
1. handle serialization better in the future (As we will know at the executor level which packages are remote installed)
2. be able to just pass the lambdas a straight-up s3 URL and not have them worry about randomly selecting which one to use, and thus they will never have to hit `meta.json`

shivaram:
@ericmjonas Your comment totally makes sense. Updated now

shivaram:
Also I'll add the module check in a separate PR. I still need to understand the cloudpickle code a bit more

pywren/wren.py
```diff
@@ -68,7 +70,9 @@ def remote_executor(config= None, job_max_runtime=3600,
     S3_PREFIX = config['s3']['pywren_prefix']
     invoker = invokers.SQSInvoker(AWS_REGION, SQS_QUEUE)
     return Executor(AWS_REGION, S3_BUCKET, S3_PREFIX, invoker, config,
-                    job_max_runtime, shard_runtime=shard_runtime)
+                    job_max_runtime)
+
+standalone_executor = remote_executor
```
shivaram: is this line necessary ? just checking as it wasn't in the original PR if i'm right

pywren/wren.py
```diff
@@ -6,6 +6,9 @@
     import pickle
 from tblib import pickling_support
 import logging
+import botocore
```
shivaram: are these imports related to this PR ?

pywren/wren.py
ooq: I think I removed it during refactoring.
Piggybacking on your PR to add it back...(bad practice)

pywren/wren.py
ooq: No. Don't know how it shows up. Now almost every file in the project has unused imports. I'm planning to get rid off them in the PyLint PR, which is coming...very soon.

Issue #16, opened 2016-10-11
Create interactive getting started script


ericmjonas:
Create a script for new users like Evan suggested, like 

```
pywren init
```
1. validate Boto config / ask them to type in aws keys and create their boto file for them
2. prompt for an S3 bucket or create an empty one for them
3. set up all the deploy scripts etc.

shivaram:
Lets fix #55 once we do the refactor to create interactive scripts

PR #89, opened 2017-03-13
Interactive setup script (Issue #16)

ericmjonas:
This is the beginning of fixing Issue  #16 . I am creating this PR now with the hope that we can get some version of it merged into master before the user camp. 

Things I have found along the way: 

1. we actually up till now had provided no way to really specify alternative config files correctly
2. at various places the AWS region was hard-coded in (!) 

So we've changed some default config variables and made it such that the command-line runner `setupcli.py` now also takes a `--filename` argument (as well as correctly respecting environment variables, etc.) 

image

I still need to write tests for this, which will exercise various edge cases, and try different AZs.


ooq:
`pytest -v tests/test_cmdline.py --runcmdtest` is the way to run tests locally?

ooq:
Nevermind. I got it.

ooq:
Maybe add `setup` as a command for `pywren`? Internally we can call `pywren-setup`. It would be nice to have all command lines to be under `pywren`.

ooq:
A bunch of typo issues: capitalize the first letter in all messages; messages should end with some punctuation marks; we should be consistent with capitalization of `PyWren`

ooq:
NIT: remove `TEST` in all default names? This feels unreliable.

ooq:
Otherwise LGTM. We can merge it now and improve it in following PRs. It's a good feature that we want ASAP.

ericmjonas:
@ooq There is still some parsing / variable name checking I think we should add before we merge, and you're right, the error messages need some more uniformity. Also the test coverage is currently abysmal.

ericmjonas:
@ooq I've fixed most of these, but could you explain what you mean by "TEST" in all default names?

ooq:
@ericmjonas Oh this is minor thing. I noticed that the lambda function names generated by default setting, for example, is called `pywren_1SCRIPT_TEST_0`. It might be confusing for certain users to view it as for testing purposes, if not all unlikely. The same goes to other names.

ooq:
I guess I just don't like the look formality of `pywren_1SCRIPT_TEST_0`...

PR #92, opened 2017-03-14
Bug fix in s3util.

ooq:


Vaishaal:
lgtm. exactly what i thought problem was :)

PR #93, opened 2017-03-14
Use S3 paginators API.

ooq:
Better API to list S3 objects.


ericmjonas:
@ooq I think these tests are failing, did we miss a merge / rebase somehow?

ooq:
@ericmjonas Just a bug in the PR. I had another commit to fix it.

PR #97, opened 2017-03-18
Refactor uses of config in executor.

### LINKED ISSUE #86: Standardize configuration for executors

shivaram:
This PR removes the config variable from the Executor class and makes the config parsing uniform across executor creations. Closes #86

### DIFF
--- a/pywren/executor.py
+++ b/pywren/executor.py
@@ -32,22 +32,21 @@ class Executor(object):
     """
 
     def __init__(self, aws_region, s3_bucket, s3_prefix,
-                 invoker, config, job_max_runtime):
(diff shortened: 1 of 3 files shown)


shivaram:
@ericmjonas I'm not sure if you had anything more in mind for #86. Right now at least the code flow for all executors is:
1. Optionally take a config input. If not use `default_config()` to parse the config file.
2. Extract the required variables from `config` and pass that into `Executor` constructor. Some of this code is shared, but I dont think its worth pulling things into a function yet.
3. Executor itself is unaware of `config`

ericmjonas:
+1

Issue #83, opened 2017-03-09
pywren not working for numba functions in 0.1 due to module serialization weirdness


Vaishaal:
```
import pywren
from numba import jit

pwex = pywren.default_executor()

def foo(x):
    return x**2

@jit
def bar(x):
    return x**2

#passes
results = pwex.map(foo, [2])
pywren.wait(results)
print(results[0].result())


#fails
results = pwex.map(bar, [2])
pywren.wait(results)
print(results[0].result())




```

This is related to #82 as this is where the nonascii.py gets imported. But it shouldn't come  into the world in the first place!

IIRC this usecase worked pre-0.1

Exception:

> Traceback (most recent call last):
>   File "numba_test.py", line 20, in <module>
>     results = pwex.map(bar, [2])
>   File "/data/vaishaal/anaconda2/lib/python2.7/site-packages/pywren/wren.py", line 265, in map
>     module_data = self.create_mod_data(mod_paths)
>   File "/data/vaishaal/anaconda2/lib/python2.7/site-packages/pywren/wren.py", line 125, in create_mod_data
>     module_data[f[len(pkg_root)+1:]] = mod_str.decode('utf-8')
>   File "/data/vaishaal/anaconda2/lib/python2.7/encodings/utf_8.py", line 16, in decode
>     return codecs.utf_8_decode(input, errors, True)
> UnicodeDecodeError: 'utf8' codec can't decode byte 0xb1 in position 81: invalid start byte

module search path for foo:
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/html/__init__.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/html/parser.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/html/entities.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/watchtower/__init__.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/http/client.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/http/cookiejar.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/http/__init__.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/http/cookies.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/http/server.py
> File name is /data/vaishaal/anaconda2/lib/python2.7/site-packages/_thread/__init__.py
[...]

Issue #82, opened 2017-03-09
pywren not respecting non UTF-8 encoding (if coding is provided at top of file)


Vaishaal:
For example if you are trying to serialize the following module:
https://github.com/ipython/ipython/blob/master/IPython/core/tests/nonascii.py

We get a unicode exception because it is not valid UTF-8

Issue #79, opened 2017-03-01
Only *.py in modules are uploaded


ooq:
This python program (https://gist.github.com/ooq/48c23aec420c682359d6a43cfc9cdd82) of mine throws `Exception: [Errno 2] No such file or directory` on remote but runs fine locally. Also it was working yesterday (was the runtime updated?) I suspect this could be a runtime issue because my local runtime uses a different boto3 botocore version. 

I'm going to debug more today by trying these two things:
1) upload my runtime
2) pull the remote runtime and setup locally

ooq:
Okay, problem identified. `Boto3` will automatically detect `certifi` and uses it for (maybe) some server authentication. This means that if `certifi` is there in my local env, it will get copied to remote as a dependent module. However, `certifi` contains a bunch of `.pem` files we are ignoring, as we only copy `.py` files:

`files = glob2.glob(os.path.join(m, "**/*.py"))`

This is why I'm getting `No such file or directory` error.

@ericmjonas

ooq:
I'll push a quick fix that puts certifi on the preinstalled list.
We should rebuild the runtimes to include certifi (more security always seems better)

ooq:
For:
`files = glob2.glob(os.path.join(m, "**/*.py"))`
A better approach is to include all files except the `*.pyc` files?

Issue #78, opened 2017-03-01
Handle exception pickling


ericmjonas:
See the comments on PR #47 at the end -- @ooq is still seeing unpickleable exceptions.

ericmjonas:
@ooq We were referencing an exception that you had seen that couldn't be serialized and yet still had a problem -- do you have a testcase perhaps ? I can't find the original problem.

ooq:
@ericmjonas Sorry that I could not recall the exact exception I had.
Just curious, do we have a rough list of 1) what exceptions cannot be pickled? 2) what exceptions can be pickled but cannot be unpickled? 
And, is `Exception` the only object that causes pickle/unpickle issues? Just want to make sure we don't have to stress out about `exc_traceback`. If `exc_traceback` is not the worry, why are we doing `str(exc_traceback)` here?

ericmjonas:
Sometimes `exc_traceback` contains pieces of the exception that cannot be pickled and I wanted to be 100% sure things propagated correctly. I don't have a list, but we have example exceptions that I test in `test_pickle.py` where we create objects (and exceptions) that fail on either pickle or unpickle.

PR #99, opened 2017-03-21
Serialization bugs

ericmjonas:
This is in aggregate an attempt to fix serialization bugs. In particular we are fixing:
#83 (@vaishaal claims it works for him)
#82 by switching to base64-encoding files we upload as part of func
#44 by using `pkgutil` to get a list of preprocessed modules as part of the runtime

In addition we no longer have a test case for  #78 so I don't know what to do there. 

At the moment the only serialization bug remaining open is #79 which is actually quite difficult and I think we should punt until subsequent releases. 

Note also that this requires an update to runtimes, which will appear in a subsequent PR.


shivaram:
Just took a quick look - most of the changes look good. I guess the remaining tricky part is generating the `preinstalls` in `runtimes` and we would need to merge that before this. 

Also I noticed you have commented out one of the test cases -- will it be re-enabled after the `runtimes` change ?

Issue #98, opened 2017-03-20
When creating the lambda we indiscriminately upload too much


ericmjonas:
PyWren's lambda indiscriminately uploads the entire source dir. If for some reason the user has put crap in there, such as a copy of the runtime, it gets uploaded as well and then the Lambda API errors telling us that the files are too big. It's really confusing. We should check that we are only uploading necessary files.

ooq:
Are you talking about users putting crap in python module folders?

ericmjonas:
Yes, exactly, but it's not that bizarre of a use case: 
1. user clones pywren github
2. user installs via `pip -e`
3.  User adds things to pywren source dir, not thinking about it. Maybe they download a runtime, as happened in this case
4. Our default upload-all policy... uploads everything, confusing them. 

I think the right answer here is a whitelist, which is easy to implement.

PR #104, opened 2017-03-23
Fix issue 98

ericmjonas:
Fix for issue #98 by whitelisting files we upload

### DIFF
--- a/pywren/scripts/pywrencli.py
+++ b/pywren/scripts/pywrencli.py
@@ -204,13 +204,16 @@ def deploy_lambda(ctx, update_if_exists = True):
     zipfile_obj = zipfile.ZipFile(file_like_object, mode='w')
 
     # FIXME see if role exists
-    files = glob2.glob(os.path.join(SOURCE_DIR, "../**/*.py"))
-    for f in files:
+    module_dir = os.path.join(SOURCE_DIR, "../")
+    
+    for f in ['wrenutil.py', 'wrenconfig.py', 's3util.py', 'wrenhandler.py', 
+              'version.py', 'jobrunner.py', 'wren.py']:
+        f = os.path.abspath(os.path.join(module_dir, f))
         a = os.path.relpath(f, SOURCE_DIR + "/..")
                             
         zipfile_obj.write(f, arcname=a)
     zipfile_obj.close()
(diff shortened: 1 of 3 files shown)


pywren/scripts/pywrencli.py
```diff
@@ -204,8 +205,11 @@ def deploy_lambda(ctx, update_if_exists = True):
     zipfile_obj = zipfile.ZipFile(file_like_object, mode='w')
 
     # FIXME see if role exists
-    files = glob2.glob(os.path.join(SOURCE_DIR, "../**/*.py"))
-    for f in files:
+    module_dir = os.path.join(SOURCE_DIR, "../")
+    
+    for f in ['wrenutil.py', 'wrenconfig.py', 's3util.py', 'wrenhandler.py', 
+              'version.py', 'jobrunner.py', 'wren.py']:
```
ooq: Is `wren.py` needed for remote side?

ooq: I feel it could be nice to still compare with files in the directory and output a warning message if there is a mismatch. Without which we might leave some hard-to-detect bugs when we add/remove files in the future.

ooq:
Do we want to do the same in `create_mod_data`?

Also I notice that `pywren` will almost always be included as a module dependency even the user function does not use it.  This might be something we want to look into.

tests/test_simple.py
```diff
@@ -318,3 +318,23 @@ def test_func(x):
...
+
+        def run_command(x):
+            return subprocess.check_output(x, shell=True).decode('ascii')
+
+        cmd = "conda info"
+        wrenexec = pywren.default_executor()
+        fut = wrenexec.call_async(run_command, cmd)
+
+        res = fut.result() 
+        assert "Current conda install" in res
```
ooq: Is this good enough? Should we also check the python version, etc.

ooq: Oh, I just notice that test is for the other PR.

pywren/scripts/pywrencli.py
ericmjonas: I don't think they'll be that hard to fix because we'll get an import error when we deploy, right?

PR #113, opened 2017-04-04
Fix issue with bucket creation in us-east-1

Donohue:
Signed-off-by: Brian Donohue <briandonohue@pinterest.com>

### DIFF
--- a/pywren/scripts/pywrencli.py
+++ b/pywren/scripts/pywrencli.py
@@ -160,9 +160,10 @@ def create_bucket(ctx):
 
     s3 = boto3.client("s3")
     region = config['account']['aws_region']
-    s3.create_bucket(Bucket=config['s3']['bucket'], 
-                     CreateBucketConfiguration={
-                         'LocationConstraint': region})
+    kwargs = {'CreateBucketConfiguration': {'LocationConstraint': region}}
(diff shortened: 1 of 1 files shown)


ooq:
@Donohue Thanks for the PR.
Can you provide some descriptions on the issue you are fixing here?

ericmjonas:
Hey Brian thanks so much for the PR -- does us-east-1 have different bucket creation semantics? @ooq we should probably run our CI in multiple regions to make sure we don't have these bugs.

Donohue:
Hey @ericmjonas and @ooq this ticket describes the issue in detail: https://github.com/boto/boto3/issues/125

Here's the relevant comment for convenience:
> This is an artifact of the underlying S3 API and the fact that it doesn't accept us-east-1. To create an S3 bucket in us-east-1 you can just not specify any CreateBucketConfiguration: s3.create_bucket(Bucket='foo').

There was an error creating the bucket in us-east-1 during setup and I verified that this resolves it.

ooq:
@Donohue Indeed! I just tried and verified this as well.
I'm surprised that they haven't fixed it after all this time.

We'll merge your PR after fixing our Travis build. Thanks again.

Donohue:
Cool 😎

PR #115, opened 2017-04-07
external shared s3 client

ericmjonas:
Improve result retrieval performance. Benchmarks forthcoming.

### DIFF
--- a/pywren/future.py
+++ b/pywren/future.py
@@ -73,7 +73,8 @@ class ResponseFuture(object):
         return True
 
 
-    def result(self, timeout=None, check_only=False, throw_except=True):
+    def result(self, timeout=None, check_only=False, throw_except=True, 
+               s3_client=None):
         """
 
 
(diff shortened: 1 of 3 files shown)


ericmjonas:
When we execute the following simple task: 
```
import pywren
import time
import cPickle as pickle
import boto3

def sleep_test(x):
    time.sleep(x)
    return x + 1

if __name__ == "__main__":


    N = 1024
    SLEEP_DUR = 100


    for use_ext_s3 in [True, False]:
        wrenexec = pywren.default_executor()
        futures = wrenexec.map(sleep_test, [SLEEP_DUR] * N)


        if use_ext_s3:
            s3_client = boto3.client('s3', 'us-west-2')
        else:
            s3_client = None

        pywren.wait(futures, s3_client=s3_client)

        results = [f.result() for f in futures]
        run_statuses = [f.run_status for f in futures]
        invoke_statuses = [f.invoke_status for f in futures]


        outdict = {'results' : results, 
                   'run_statuses' : run_statuses, 
                   'invoke_statuses' : invoke_statuses, 
                   'N' : N, 
                   'SLEEP_DUR' : SLEEP_DUR}
        pickle.dump(outdict, open("invocation.{}.{}.pickle".format(N, use_ext_s3), 'w'))

```
This is the behavior that is current in master, where we create new s3 clients inside each query: 
image

and this is the behavior when we share them [`use_ext_s3=True` in the above code]: 

image

PR #117, opened 2017-04-08
Added validation for s3 bucket names and made username portion of def…

sean-smith:
…ault bucket name lowercase


ooq:
Thanks for the PR @sean-smith ! LGTM.

Issue #109, opened 2017-03-29
S3 connection pooling and client creation / lifecycle 


ericmjonas:
In branch https://github.com/pywren/pywren/tree/debug_result_performance I have moved the s3 client invocation around so that we can set up a single boto s3 object. If we pass `s3_client=None` then the current (master branch) behavior is run. 

currently this is the code path we run
```
s3_client = None
 pywren.wait(futures, WAIT_DUR_SEC=1, s3_client=s3_client)
```

and we see:
![image](https://cloud.githubusercontent.com/assets/36626/24469903/f03878f4-1471-11e7-9c48-99478c21f061.png)

vs when we create the client external to `wait` and share it across threads:
```
s3_client = boto3.client('s3', 'us-west-2')
pywren.wait(futures, WAIT_DUR_SEC=1, s3_client=s3_client)
```

![image](https://cloud.githubusercontent.com/assets/36626/24469929/07586990-1472-11e7-90d8-e7b39460acc6.png)

I think we should dig into how boto connection pooling actually works and where our time is going when we retrieve these results. I think this might also inform our current storage refactoring #108

ooq:
I think this issue should be resolved as a result of storage refactoring, when we hide s3 client objects under the storage layer.
The next problem with reusing one S3 client is that the default connection pool size is 10, which is far beyond enough. So we either have to monkey-patch botocore, or we need to maintain a pool of clients.

ericmjonas:
@ooq have you tried this? 
```
s3_client = boto3.client('s3', 'us-west-2'
                        config=botocore.client.Config(max_pool_connections=50))
```

ooq:
@ericmjonas I haven't! This should work perfectly.

ooq:
@Vaishaal Likely would solve your "connection pool" problem too.

Issue #23, opened 2016-10-23
Consider using boto3 s3 transfer interface


ericmjonas:
http://boto3.readthedocs.io/en/latest/_modules/boto3/s3/transfer.html

shivaram:
This will be a part of the discussion in #108

ooq:
So the transfer APIs are only for files. I think the only place we can use it now is in wrenhandler.py. I'll change the wrenhandler to use transfer API.

PR #119, opened 2017-04-09
[Issue#108] Refactor Storage API

ooq:
This PR:
1, creates a Storage class as a middle layer to separate PyWren core and S3.
2, creates S3Service that wraps S3 boto3 APIs. In S3Service, reuses the same S3Client per #109 .
3, adds `pywren. save_futures_to_string`, `pywren.load_futures_from_string`, `pywren.save_futures_to_file`, `pywren.load_futures_from_file`. 
4, in `wrenhandler.py`, use #23 .
5, removes `s3util.py`.
6, configuration: each object/method should only be exposed to configuration that is needed for that object/method. For example, the config file used to create `Storage` should be generated by `wrenconfig.extract_storage_config(config)`; `config['runtime']` should be passed to `runtime.get_runtime_info(config['runtime'])` instead of `config`.
7, documentation is added for new classes and methods.
8, no new test is added.

@shivaram @ericmjonas


ooq:
Okay, I'll make changes per discussion with @ericmjonas yesterday.

Just for record, below is a performance graph for #109 
image

ooq:
Done. 
image
#119 should be resolved.

pywren/future.py
```diff
@@ -103,10 +100,10 @@ def result(self, timeout=None, check_only=False, throw_except=True):
             else:
                 return None
 
+        if storage_handler is None:
```
shivaram: Did we reach consensus on what the model we want to support is ? There was the other option of passing in the storage_handler while constructing the future that I think avoids some of the failure scenarios ?

pywren/runtime.py
```diff
@@ -2,24 +2,32 @@
 import json
 import sys
 
-def get_runtime_info(bucket, key):
+import pywren.storage as storage
+import pywren.wrenconfig as wrenconfig
+
+# FIXME separate runtime code with S3
```
shivaram: Can you clarify what this fixme means ?

pywren/runtime.py
```diff
@@ -2,24 +2,32 @@
...
+
+def get_runtime_info(runtime_config, storage_handler = None):
     """
     Download runtime information from S3 at deserialize
     """
-    s3 = boto3.resource('s3')
+    if storage_handler is None:
+        storage_config = wrenconfig.extract_storage_config(wrenconfig.default())
+        storage_handler = storage.Storage(storage_config)
```
shivaram: Similar to above, is there a case where we can't pass in a `Storage` object ?

pywren/storage/s3_service.py
```diff
@@ -0,0 +1,87 @@
...
+        :param data: data of the object
+        :return: None
+        """
+        self.s3client.put_object(Bucket=self.s3_bucket, Key=key, Body=data)
+
+    def get_object(self, key):
+        """
+        Get object from S3 with a key.
+        :param key: key of the object
+        :return: Data of the object
```
shivaram: Can we specify the type of the return object ? i.e. is it a byte array or an iterator etc ?

pywren/storage/storage.py
```diff
@@ -0,0 +1,140 @@
...
+        return self.service_handler.get_object(key)
+
+    def create_keys(self, callset_id, call_id):
+        """
+        Create keys for data, output and status given callset and call IDs.
+        :param callset_id: callset's ID
+        :param call_id: call's ID
+        :return: data_key, output_key, status_key
+        """
+        data_key = os.path.join(self.prefix, callset_id, call_id, "data.pickle")
```
shivaram: all of these suffixes are hardcoded right now. This is probably not a problem for this PR, but we should come back and refactor this

pywren/wren.py
```diff
@@ -71,17 +60,50 @@ def remote_executor(config=None, job_max_runtime=3600):
...
     invoker = invokers.SQSInvoker(AWS_REGION, SQS_QUEUE)
-    return Executor(AWS_REGION, S3_BUCKET, S3_PREFIX, invoker,
-                    RUNTIME_S3_BUCKET, RUNTIME_S3_KEY, job_max_runtime)
+
+    return Executor(invoker, config, job_max_runtime)
 
 standalone_executor = remote_executor
 
 
+def save_futures_to_string(futures):
```
shivaram: are we adding these functions in this PR ? Not sure it relates to the storage refactoring ? Can we move this out and discuss this separately ?

pywren/wrenhandler.py
```diff
@@ -37,7 +36,17 @@
 
 PROCESS_STDOUT_SLEEP_SECS = 2
 
-def download_runtime_if_necessary(s3conn, runtime_s3_bucket, runtime_s3_key):
+def get_key_size(s3client, bucket, key):
```
shivaram: just to clarify this is s3 specific right now ?

pywren/future.py
ooq: Yes, we (Eric and I) agreed that we want futures to be serializable. Storing storage handler as a field of future will make it un-serializable, my previous solution was to have a `save()` method in which we first set `future.storage_handler=None` to make future serializable. We then agree that having `pickle.dump(futures)` to work is also important.

pywren/runtime.py
ooq: Oh, this has been fixed. Will remove.

pywren/runtime.py
ooq: This can happen in tests, etc.

pywren/storage/s3_service.py
ooq: Sure.

pywren/storage/storage.py
ooq: ditto

pywren/wren.py
ooq: makes sense.

pywren/wrenhandler.py
ooq: yes.

ooq:
@shivaram @ericmjonas Comments addressed.

pywren/executor.py
```diff
@@ -85,18 +74,17 @@ def invoke_with_keys(self, s3_func_key, s3_data_key, s3_output_key,
             random.seed()
             runtime_url = random.choice(self.runtime_meta_info['urls'])
 
-
-        arg_dict = {'func_key' : s3_func_key,
-                    'data_key' : s3_data_key,
-                    'output_key' : s3_output_key,
-                    'status_key' : s3_status_key,
+        arg_dict = {'storage_info' : self.storage.get_storage_info(),
```
shivaram: This is the main thing that we discussed the other day -- can we see if there a better way to serialize / send information from the driver to the workers ? One possibility could be that we send a `config` object or the result of `extract_storage_config` and then recreate the StorageHandler on the worker side ?

pywren/executor.py
```diff
@@ -181,19 +167,16 @@ def map(self, func, iterdata, extra_env = None, extra_meta = None,
...
-        s3_agg_data_key = None
-        host_job_meta['aggregated_data_in_s3'] = False
+        agg_data_key = None
+        host_job_meta['agg_data'] = False
         host_job_meta['data_size_bytes'] =  data_size_bytes
 
         if data_size_bytes < wrenconfig.MAX_AGG_DATA_SIZE and data_all_as_one:
-            s3_agg_data_key = s3util.create_agg_data_key(self.s3_bucket,
-                                                         self.s3_prefix, callset_id)
+            agg_data_key = self.storage.create_agg_data_key(callset_id)
```
shivaram: so one more thought -- all these utility functions to create keys, can we stick them into a `StorageUtils` or something like that ? These functions shouldn't change based on storage backend, is that right ?

pywren/storage/storage.py
```diff
@@ -0,0 +1,140 @@
...
+        output_key = os.path.join(self.prefix, callset_id, call_id, "output.pickle")
+        status_key = os.path.join(self.prefix, callset_id, call_id, "status.json")
+        return data_key, output_key, status_key
+
+    def create_func_key(self, callset_id):
```
shivaram: This is related to the previous comment -- I feel like we should only have the abstract methods in this class so it is very clear what needs to be implemented for a new implementation ?

pywren/wrenhandler.py
```diff
@@ -46,7 +55,7 @@ def download_runtime_if_necessary(s3conn, runtime_s3_bucket, runtime_s3_key):
     """
 
     # get runtime etag
-    runtime_meta = s3conn.meta.client.head_object(Bucket=runtime_s3_bucket, 
+    runtime_meta = s3_client.head_object(Bucket=runtime_s3_bucket,
```
shivaram: It will be a good thing to think about what parts of wrenhandler cannot be handled by our existing storage API. I guess right now this includes `head_object`, `download_file` and `upload_file` ?

pywren/storage/storage.py
```diff
@@ -0,0 +1,140 @@
...
+    def create_agg_data_key(self, callset_id):
+        """
+        Create aggregate data key
+        :param callset_id: callset's ID
+        :return: a key for aggregate data
+        """
+        agg_data_key = os.path.join(self.prefix, callset_id, "aggdata.pickle")
+        return agg_data_key
+
+    def get_callset_status(self, callset_id):
```
shivaram: This should probably be renamed. It doesn't really return the status right ? I'd call it something like `get_calls_in_callset` ?

shivaram:
@ooq Can you rebase with the latest fix ?

pywren/wrenhandler.py
shivaram: Did we reach any longer term consensus on this ? If not we can talk about it this week

pywren/storage/storage.py
```diff
@@ -0,0 +1,111 @@
...
+    Currently we only support S3 as the underlying service.
+    We plan to support other services in the future.
+    """
+
+    def __init__(self, config):
+        self.storage_config = config
+        self.prefix = config['storage_prefix']
+        self.service = config['storage_service']
+        if config['storage_service'] == 's3':
+            self.service_handler = S3Service(config['service_config'])
```
shivaram: can we call this storage backend ? might be better than service_handler. Also it might be better to have an explicit interface for storage backend - that makes it easier to code / test ?

pywren/executor.py
```diff
@@ -32,50 +36,39 @@ class Executor(object):
...
                          callset_id, call_id, extra_env,
                          extra_meta, data_byte_range, use_cached_runtime,
                          host_job_meta, job_max_runtime,
                          overwrite_invoke_args = None):
 
-        # Pick a runtime url if we have shards. If not the handler will construct it
-        # using s3_bucket and s3_key
+        # Pick a runtime url if we have shards.
+        # If not the handler will construct it
+        # TODO: we should always set the url, so our code here is S3-independent
```
ooq: This TODO is actually fixed. Removing this comment.

pywren/executor.py
```diff
@@ -2,6 +2,9 @@
 
 import boto3
 import json
+
+from pywren.storage.storage_utils import create_func_key
```
ooq: Let's do this in style check.

pywren/storage/storage.py
ooq: StorageBackend sounds better. 
I looked into the python way for interface , and Abstract Base Class, which seems to be the to-go method. The benefit of having an interface class for code/test is probably very limited for a dynamic language, as we probably won't get NotImplementedError from IDE or install time. To that end, this is much worth to add more non-pythonic code?

ooq:
Thanks for the comments @shivaram . The PR is updated.

ooq:
@apengwin Can you also take a look at this?

pywren/executor.py
```diff
@@ -32,50 +36,38 @@ class Executor(object):
...
-        self.s3client = self.session.create_client('s3', region_name = aws_region)
         self.job_max_runtime = job_max_runtime
 
-        self.runtime_bucket = runtime_s3_bucket
-        self.runtime_key = runtime_s3_key
-        self.runtime_meta_info = runtime.get_runtime_info(runtime_s3_bucket, runtime_s3_key)
-        if not runtime.runtime_key_valid(self.runtime_meta_info):
-            raise Exception("The indicated runtime: s3://{}/{} is not approprite for this python version".format(runtime_s3_bucket, runtime_s3_key))
+        self.config = config
+        self.storage_config = wrenconfig.extract_storage_config(self.config)
```
apengwin: Can we formalize what `config` looks like? Also, it might be easier to have just one `self.config` instead of both `self.storage_config` and `self.config`

ooq: I have wanted to formalize/document config for quite some time. For now, this is structured/parsed in a way to be backward-compatible with .pywren_config. Let's formalize config in a separate PR. 
For the `self.storage_config`, so the idea is to make the config more "hierarchical" and only expose parameters that are related to class.

tests/test_queues.py
```diff
@@ -17,6 +17,8 @@ class SimpleAsync(unittest.TestCase):
     Test sqs dispatch but with local runner
     """
     def setUp(self):
```
apengwin: Should this be `set_up`

ooq: We do have some inconsistent formatting now and needs to be fixed.
Can the purpose of this PR, can you just review the changes?

pywren/executor.py
apengwin: This is kind of petty, but it's confusing to have multiple `config` variables that are all different, so should we change to something like `storage_info`

pywren/executor.py
```diff
@@ -135,9 +126,8 @@ def invoke_with_keys(self, s3_func_key, s3_data_key, s3_output_key,
 
         host_job_meta.update(arg_dict)
 
-        fut = ResponseFuture(call_id, callset_id, host_job_meta,
-                             self.s3_bucket, self.s3_prefix,
-                             self.aws_region)
+        storage_path = storage_utils.get_storage_path(self.storage_config)
```
shivaram: nit: This can probably be done at the beginning where we construct storage_config ?

pywren/executor.py
```diff
@@ -176,27 +166,25 @@ def map(self, func, iterdata, extra_env = None, extra_meta = None,
         host_job_meta = {}
 
         pool = ThreadPool(invoke_pool_threads)
-        callset_id = s3util.create_callset_id()
+        callset_id = wrenutil.create_callset_id()
+        data = list(iterdata)
```
shivaram: Strange that this shows up in the diff - how was it working before without this line ?

apengwin: `s3util.create_callset_id` is just a wrapper around `wrenutil.create_callset_id`

apengwin: The `data=list(iterdata)` line does appear twice here. @ooq ?

screen shot 2017-06-30 at 11 00 16 pm

PR #121, opened 2017-04-10
Add utility to create ssh key; add it to travis test.

ooq:
When I setup Travis on my fork repository, Standalone and Macreduce mode will fail because ssh key `ec2-us-west-2` does not exist.
This patch 1) adds an utility to create ssh key; 2) in Travis tests, creates the ssh key if it does not already exist.

### DIFF
--- a/pywren/scripts/pywrencli.py
+++ b/pywren/scripts/pywrencli.py
@@ -159,7 +159,8 @@ def create_ssh_key(ctx):
     config_filename = ctx.obj['config_filename']
     config = pywren.wrenconfig.load(config_filename)
 
-    ec2 = boto3.client('ec2')
+    region = config['account']['aws_region']
+    ec2 = boto3.client('ec2', region)
     keyname = config['standalone']['ec2_ssh_key']
     try:
         ec2.describe_key_pairs(KeyNames = [keyname])
--- a/pywren/scripts/pywrencli.py
+++ b/pywren/scripts/pywrencli.py
@@ -129,7 +129,7 @@ def test_config(ctx):
 @click.pass_context
 def create_role(ctx):
     """
-    
+
     """
     config_filename = ctx.obj['config_filename']
     config = pywren.wrenconfig.load(config_filename)
@@ -137,10 +137,10 @@ def create_role(ctx):
     iamclient = boto3.resource('iam')
     json_policy = json.dumps(pywren.wrenconfig.basic_role_policy)
     role_name = config['account']['aws_lambda_role']
-    role = iamclient.create_role(RoleName=role_name, 
+    role = iamclient.create_role(RoleName=role_name,
                                  AssumeRolePolicyDocument=json_policy)
     more_json_policy = json.dumps(pywren.wrenconfig.more_permissions_policy)
-    
+
     AWS_ACCOUNT_ID = config['account']['aws_account_id']
     AWS_REGION = config['account']['aws_region']
     more_json_policy = more_json_policy.replace("AWS_ACCOUNT_ID", str(AWS_ACCOUNT_ID))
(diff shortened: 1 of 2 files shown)


shivaram:
Does this solve the full problem with running travis on forks ? Or is this one of the steps required ?

ooq:
This should solve the full problem. @shivaram

pywren/scripts/pywrencli.py
```diff
@@ -149,6 +149,26 @@ def create_role(ctx):
...
+    config = pywren.wrenconfig.load(config_filename)
+
+    region = config['account']['aws_region']
+    ec2 = boto3.client('ec2', region)
+    keyname = config['standalone']['ec2_ssh_key']
+    try:
+        ec2.describe_key_pairs(KeyNames = [keyname])
+    except botocore.exceptions.ClientError as e:
+        if e.response['Error']['Code'] == 'InvalidKeyPair.NotFound':
+            ec2.create_key_pair(KeyName = keyname)
```
shivaram: Just curious - where does the private key for this keypair go to ? Is it downloaded to some known location ?

ooq: It is not downloaded, and for the purpose of the tests, the key is not required.

shivaram: I see - if we add it as a CLI command then it could be also used by new users who dont know the AWS interface etc. So lets just save it to some (user customizable) location if its not that hard to do it ?

ooq:
@shivaram I added the save. The user can pass a location to the function and save the ssh key. (otherwise it's not saved for security concerns)

ericmjonas:
I agree this is useful for standalone users, can we add it to the interactive script as well? 

@shivaram Can you expand on what you mean with "full problem when running travis on forks"? This will let users with forks set up their own travis, but it will not solve the pull-request-travis problem.

shivaram:
Yeah that is what I was thinking of. I guess this is useful irrespective of not solving the "pull-request-on-travis" problem

shivaram:
LGTM

shivaram:
@ooq Any idea why travis is failing ?

ooq:
@shivaram Travis says unauthorized permission when creating keys.
@ericmjonas Is it possible that the AWS credentials in Travis provide restricted permissions?

ericmjonas:
@ooq yeah I totally bet that is it. Do you have the error message?

ericmjonas:
That will tell me the permission(s) to grant the IAM role

ooq:
@ericmjonas 
`botocore.exceptions.ClientError: An error occurred (UnauthorizedOperation) when calling the DescribeKeyPairs operation: You are not authorized to perform this operation.`
That is `DescribeKeyPairs`, you might also need to add `CreateKeyPair`

ericmjonas:
Ok I've added
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Stmt1493213587000",
            "Effect": "Allow",
            "Action": [
                "ec2:CreateKeyPair",
                "ec2:DeleteKeyPair",
                "ec2:DescribeKeyPairs"
            ],
            "Resource": [
                "*"
            ]
        }
    ]
}
```
to the user, can we see if that works?

ooq:
hmm, still failing.

shivaram:
@ericmjonas is there a way to print the available permissions for the role during the travis test ? Right now we run `pywren test_config` and `pywren get_aws_account_id`

ooq:
The error is still ``botocore.exceptions.ClientError: An error occurred (UnauthorizedOperation) when calling the DescribeKeyPairs operation: You are not authorized to perform this operation.''

Should we address this after we migrate the Travis tests to a dedicated PyWren account?
It seems hard to debug this problem now. @shivaram @ericmjonas

shivaram:
In that case we should remove the failing test case for now ?

PR #132, opened 2017-04-28
Add Apache v2 License

### LINKED ISSUE #116: Create License File
I'm happy with a 3-clause BSD or MIT, I don't see any reason we should be restrictive. Thoughts?

shivaram:
Closes #116


shivaram:
Thanks @ooq !

PR #133, opened 2017-04-28
Handle empty input list in map

### LINKED ISSUE #129: Mapping over an empty list should return an empty list
This is super dumb but mapping over an empty list shouldn't throw an exception. It should just return an empty list.

`pwex.map(lambda x: x, [])`

> ---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-171-c909ffd0669a> in <module>()
----> 1 pwex.map(lambda x: x, [])

/home/ubuntu/pywren/pywren/executor.py in map(self, func, iterdata, extra_env, extra_meta, invoke_pool_threads, data_all_as_one, use_cached_runtime, overwrite_invoke_args, exclude_modules)
    281         pool.close()
    282         pool.join()
--> 283         logger.info("map invoked {} {} pool join".format(callset_id, call_id))
    284 
    285         # FIXME take advantage of the callset to return a lot of these

UnboundLocalError: local variable 'call_id' referenced before assignment

shivaram:
Fixes #129

### DIFF
--- a/pywren/executor.py
+++ b/pywren/executor.py
@@ -169,11 +169,14 @@ class Executor(object):
         to False, redownloads runtime.
         """
 
+        data = list(iterdata)
+        if not data:
+          return []
+
         host_job_meta = {}
 
         pool = ThreadPool(invoke_pool_threads)
         callset_id = s3util.create_callset_id()
-        data = list(iterdata)
 
         ### pickle func and all data (to capture module dependencies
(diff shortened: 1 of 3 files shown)


shivaram:
@ooq @ericmjonas I marked the numba test as flaky in this PR as I ran into two travis failures related to numba. Let me know if there is a better fix

Issue #136, opened 2017-05-16
pywren executors are over aggressive in uploading modules


Vaishaal:
If you import a python module to access a function inside it, ALL  the dependencies of that module are recursively imported and uploaded with the function. On storage limited lambda instances this leads you to run out of memory.

ooq:
So in most cases the recursively imported modules are not needed? Maybe an example here would be nice.

Vaishaal:
Yes its common in large-ish python projects.  Here is an example


library.py:
```
import big_module
def foo(x):
      return big_module.foo(x)

def bar(x):
       return x
```

driver.py:
```
import library 
pwex = pywren.default_executor()
pwex.map(library.bar, [0,1,2,3])
```

obviously this example is stupid. But I can link you to real examples in my project.

ooq:
I see. This example makes a lot of sense.
I don't mind adding a "exclude_module" operator which just gives user more control. But just to explore the solution space here, is it also possible to solve it in the dependency analyzer?
Maybe @ericmjonas or @shivaram who is more familiar with the serialization stuff can also chime in here.

curlyz:
I am amazed because this is a 3 years old issue.
Even running the simplest code cost my lambda memory up to 1GB, ( still don't know why ).
I suspect that pywren pack all the imported library even though it is not related to the function.
I am wondering if this trick will filter out unneeded library.
 
``` python
    def createExecutionCode(self, fn, input_value):
        if (fn, input_value) in self.executionDb:
            return self.executionDb[(fn, input_value)]

        # :: Get all involved function with this one
        list_involved_function = [fn]
        while True:
            doAnotherRound = False
            for function in list_involved_function:
                for global_name, global_value in inspect.getclosurevars(function).globals.items():
                    if inspect.isfunction(global_value):
                        if global_value not in list_involved_function:
                            list_involved_function.append(global_value)
                            doAnotherRound = True
            if not doAnotherRound:
                break

        # :: Get all the used pip library
        pip_list = {}
        for function in list_involved_function:
            for global_name, global_value in inspect.getclosurevars(function).globals.items():
                if inspect.ismodule(global_value):
                    if "site-packages" in repr(global_value):
                        pip_list[global_name] = None
                        if global_name not in self.listPipWarnedLibrary:
                            self.listPipWarnedLibrary.append(global_name)
                            logger.warning("[Pip] This is an installed library : {}, if the pip package name is different, please specify in pip_alias on Executor arguments".format(global_name))
```

PR #140, opened 2017-05-17
add exclude modules argument to map

Vaishaal:
Exclude modules in map based on prefix

Fix to #136


ooq:
Let's maybe have an example to document this?

Vaishaal:
Here is an example


library.py:
```
import big_module
def foo(x):
      return big_module.foo(x)

def bar(x):
       return x
```

driver.py:
```
import library 
pwex = pywren.default_executor()
pwex.map(library.bar, [0,1,2,3])
```

Vaishaal:
I've also added a test. @ericmjonas  could you see if this is sufficient?

tests/test_modules.py
```diff
@@ -126,4 +126,27 @@ def test_simple(self):
...
+            return extmodule.foo_add(x)
+        x = 1.0
+        fut0 = self.wrenexec.map(foo, [x])
+        res = fut0[0].result()
+        self.assertEqual(res, 2.0)
+
+        fut = self.wrenexec.map(foo, [x], exclude_modules=["extmodule"])
+        try:
+            fut[0].result()
+            self.fail("shouldn't happen")
```
ooq: If exclusion does not work, there might still be some exception because extmodule does not really exist right?

Vaishaal: What do you mean? In this case extmodule exists right?

ooq: Did you import it?

Vaishaal: at the top. line 6.

ooq: Ah okay, I missed it.

shivaram:
I think this looks fine to me. @ooq do you want to take one final look ?

ooq:
Looks good. Will merge.

PR #142, opened 2017-06-15
Remove cloudpickle dependency and use local cloudpickle.

ooq:
This PR will fix the build error on master.
Somehow we are not using the local cloudpickle and the latest cloudpickle via pip is not compatible.


shivaram:
Nice work debugging this @ooq ! LGTM. @ericmjonas Any thoughts on this ? If not lets merge this and make our builds green

ericmjonas:
LGTM

PR #143, opened 2017-07-03
Pylint cleanup

shivaram:
This PR contains changes to address warnings from Pylint. Right now this includes removing trailing whitespace and organizing the imports.


shivaram:
There are still some more warnings left. The current output of running `pylint --rcfile ./pylintrc pywren` is at https://gist.github.com/shivaram/039511355e726ea7a277f51d3032e1ac
Some of these warnings are simple to fix (like the print formatting). 

The other warnings of relative import seem more tricky to fix as from my testing we rely on relative imports to get the lambda runner to work correctly.

Also @ooq @ericmjonas it would be good if you could look at the list of disabled checks at https://github.com/pywren/pywren/blob/pylint/pylintrc#L84 and see if something should be enabled.

shivaram:
Ok I enabled a bunch of more rules and fixed issues raised in them. https://gist.github.com/shivaram/5e297f237fabd268b860f367fb2f9331 is the latest remaining errors. 

Some of the remaining ones are about unused variables which I am a bit hesitant to change in this PR as that will need more careful reviewing ?

ooq:
You have more commits on this, right? @shivaram

shivaram:
I will rebase this with @Vaishaal's recently merged PR. I have some more commits to follow that make some more code changes -- Should we do this in this same PR or do it in a follow up PR ?

ooq:
I think the idea of having a "safe" PR that requires minimal code change, and a follow up PR that finalizes lint, is pretty reasonable. We can pass the first one very quickly.

shivaram:
Great. Then I will stash the other commits. Can you take a look at the existing ones ? I'll just push one more commit with the rebase

shivaram:
Seems like a flaky test run - `Value (pywren_travis_589_7_052493) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name` - Re-triggering the tests

shivaram:
@ooq - tests pass now

pywren/wren.py
```diff
@@ -60,7 +59,7 @@ def remote_executor(config=None, job_max_runtime=3600):
 
     AWS_REGION = config['account']['aws_region']
     SQS_QUEUE = config['standalone']['sqs_queue_name']
-    invoker = invokers.SQSInvoker(AWS_REGION, SQS_QUEUE)
+    invoker = queues.SQSInvoker(AWS_REGION, SQS_QUEUE)
```
ooq: Is this caused by import changes due to lint? I feel previous code was cleaner.

shivaram: I'm not sure how this was working before. From what I can see SQSInvoker is only present in queues.py ? So was it somehow being imported by invokers.py and then re-exported ? I could try to revert this and see if lint and tests are happy, but I'm also curious what the intended behavior here is

ooq: I think the invokers should be a module and the queues.py just provide some implementations. So it's nice to access all invokers from invokers.py. 
Personally I feel having a separate`queues.py` file is not very necessary for now, but we might expand more in the future.

ooq: Does lint complain if you revert back?

shivaram: So to do this I think we need to import SQSInvoker into invokers.py ? If I do that then I get `W:  9, 0: Unused SQSInvoker imported from pywren.queues (unused-import)` in invokers.py.

If I dont do that import then I get `E: 61,14: Module 'pywren.invokers' has no 'SQSInvoker' member (no-member)`

Again -- I'm not sure having the class in queues.py is a good idea if we want to call it from invokers.py (if we need a module we need to create invokers as a directory ? that seems overkill)

ooq: Agree that would overkill.
Another way is to wrap SQSInvoker in `invokers.py` (We probably want to change the naming  as well, having a invoker inside a queue.py is not ideal). But this should not be in this PR.
I'm fine with merging.

shivaram: Ok cool. I just fixed one minor thing due to a merge conflict with vaishaal's PR. Feel free to merge once the tests pass !

PR #147, opened 2017-07-17
change docstring

apengwin:


pywren/storage/s3_backend.py
```diff
@@ -60,8 +60,8 @@ def key_exists(self, key):
     def list_keys_with_prefix(self, prefix):
         """
         Return a list of keys for the given prefix.
-        :param key: key of the object
-        :return: True if key exists, False if not exists
+        :param prefix: Prefix to filter object names.
+        :return: Objects of the bucket that match prefix
```
shivaram: It returns just the keys right and not the objects ?

apengwin: Yes.

PR #148, opened 2017-07-18
pull default from github

apengwin:
Fix a FIXME

### DIFF
--- a/pywren/scripts/pywrencli.py
+++ b/pywren/scripts/pywrencli.py
@@ -15,7 +15,7 @@ import click
 import pywren
 import pywren.runtime
 from pywren import ec2standalone
-
+import urllib2
 
 @click.group()
 @click.option('--filename', default=pywren.wrenconfig.get_default_config_filename())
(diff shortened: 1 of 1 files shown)


shivaram:
Couple of travis related things
- urllib2 isn't present in Python 3 as the errors tell you. We'll again need to add some switch case code -- Is there a better way to handle this without using urllib2 ?
- Could you run `pylint` to make sure this change doesn't introduce new errors ? The travis + pylint PR is still pending, but it might be better to not introduce new errors

apengwin:
@shivaram  Python's builtin libraries for fetching data over HTTP have been changed from python 2.7-> 3, but there's `from urllib.request import urlopen` which works the same way.

ericmjonas:
Hey guys, this entire PR confuses me; @apengwin can you provide a little bit more context for why the default yaml wouldn't exist in the install dir?

ooq:
I have the same confusion as @ericmjonas.
Is the goal here to fix the #fixme? If that's the case, it would be nice to throw some exception if the yaml file does not exist. But i'm not sure if going further to pull from github is the right way to handle this.

ooq:
@apengwin should we close this?

PR #158, opened 2017-08-15
Make sure pylint passes before running travis tests.

### LINKED ISSUE #149: Passing pylint should be a prerequisite for actual travis tests
Right now our linting tests are run in a separate mode as `RUN_PYLINT=true` (see .travis.yaml), and other modes, such as `RUN_LAMBDA=true` are run in parallel. A more ideal setup would be to run modes like `RUN_LAMBDA=true` only after `RUN_PYLINT=true` succeeds. 

This might be achieved via implementing complex build steps in travis:
https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps

### LINKED ISSUE #157: Do not proceed with travis build if pylint fails.

apengwin:
Fix #149


shivaram:
This might not be easy because pylint might not be installed and thus we can't call it beforehand. Also right now we use this environment variable from the test matrix that you'd need to change to test this.

apengwin:
Looking at the travis logs, the build does stop if pylint doesn't pass, but we're hitting a bunch of import errors since the conda env hasn't been built yet.

shivaram:
@apengwin yeah I'm not sure if this is the right approach given the import errors. Here is one more idea -- Since lint runs fast can we call it from run-tests.sh for all the cases (lambda, standalone etc.) What I'm suggesting is that we can add a block of code to `lambda_tests.sh` that first calls pylint and if that succeeds it runs `pytest`

Issue #162, opened 2017-08-20
config files manually loaded must be patched with storage config


ericmjonas:
Right now our config files don't have any storage handler information. This is fine, but when we manually load them (as opposed to calling `default()`) we need to patch them.

PR #167, opened 2017-08-22
Fix patching of storage handler config, then fix commandline

### LINKED ISSUE #166: Test Function doesn't parse command-line config
Currently test function in the set script doesn't use the command-line-supplied config, and instead uses the default. This should be fixed as test_function is a good way to test a given config/setup, esp for risecamp

ericmjonas:
Fixes #166 #162

### DIFF
--- a/pywren/wrenconfig.py
+++ b/pywren/wrenconfig.py
@@ -1,5 +1,5 @@
 import os
-
+import copy
 
 
 GENERIC_HANDLER_NAME = "wrenhandler.generic_handler"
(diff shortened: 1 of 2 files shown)


pywren/wrenconfig.py
```diff
@@ -60,6 +63,17 @@ def get_default_config_filename():
 
     return config_filename
 
+def patch_storage_config(config_data):
```
shivaram: Can we print a warning or something here ? I'm trying to understand if this is an unusual case where we need to tell the user that we are using s3 as the default storage or if this is an expected workflow

ericmjonas:
For the time being this is 100% the expected workflow, because we are not
yet migrating user configurations. In a later release we should add a
warning, telling people to upgrade their config file format.

Issue #164, opened 2017-08-21
Change default runtime URLS to be region-specific


ericmjonas:
Change the default runtime to be region-specific and put runtimes in each region.

PR #169, opened 2017-08-23
Switch to region specific runtimes

ericmjonas:
This pull request changes:
1. the defaults that we use for runtimes
2. The way we lookup runtime information

In particular, we now explicitly get the runtime meta.json instead of having the hack where we assume there is some root .tar.gz. This should have no user-visible impact; the sharded runtime has been default for quite a while. 

We now have runtimes and associated buckets in all the major US regions, in public buckets. It's easy to add more, as well. 

I am not tying this to the new runtime builder as technically they are independent. Old runtimes should work with this code as well. 

Closes issue #164

### DIFF
--- a/tests/test_simple.py
+++ b/tests/test_simple.py
@@ -305,24 +305,24 @@ class RuntimeSharding(unittest.TestCase):
     #     self.assertEqual(future.run_status['runtime_s3_key_used'],
     #                      base_runtime_key)
 
-    def test_shard(self):
-        config = pywren.wrenconfig.default()
-        old_key = config['runtime']['s3_key']
-        prefix, tar_gz = os.path.split(old_key)
-        # Use a runtime that has shards
-        config['runtime']['s3_key'] = os.path.join("pywren.runtimes", tar_gz)
-        wrenexec = pywren.default_executor(config=config)
+    # def test_shard(self):
+    #     config = pywren.wrenconfig.default()
+    #     old_key = config['runtime']['s3_key'] 
+    #     prefix, tar_gz = os.path.split(old_key)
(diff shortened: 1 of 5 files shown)


tests/test_simple.py
```diff
@@ -288,41 +288,41 @@ class RuntimeSharding(unittest.TestCase):
...
-            return x + 1
-
-        base_runtime_key = config['runtime']['s3_key']
-
-        future = wrenexec.call_async(test_func, 7)
-        result = future.result()
-        # NOTE: There is some probability we will hit the base key ? 
-        self.assertNotEqual(future.run_status['runtime_s3_key_used'], 
-                         base_runtime_key)
+    # def test_no_shard(self):
```
shivaram: just to be clear is this test not going to be applicable from here on or is it that we will add shards to the new region-specific runtimes as well ?

ericmjonas: Sorry I don't quite understand the question; All future runtimes, including those that are region-specific, will be sharded. Thus eliminating these tests. 

This seems reasonable because: 
1. if you don't shard the runtime you run into performance issues _really_ quickly
2. All "sharding" technically is is sticking the runtime at some locations pointed to by the meta.json. So we could in the future support single-tarball sharding if we wanted with no change in code. In fact, for testing, I often don't fully shard and just have the meta.json point to a single tarball.

shivaram: Got it. I was just wondering which way we went in terms of all sharding vs. no sharding. 
Do we still want to test the meta.json -> shards being picked up correctly ? If not we can just delete these tests.

ericmjonas:
Ok @shivaram got rid of that test

Issue #160, opened 2017-08-18
Straggler due to list-after-write consistency


ooq:
According to this post:
https://forums.aws.amazon.com/thread.jspa?messageID=687028#687028

S3 provides only eventual-consistency for list() operation after writes. This, according to my experience, might result in 20-30 seconds of delay for discovering the results of a finished task. One solution is to use `get` to fetch results after, say, 90% of tasks are finished.

I have a patch to fix this issue. Will make a PR.

ericmjonas:
Fixed by PR #170

PR #170, opened 2017-08-23
Issue#160 Straggler due to list-after-write consistency

ooq:
In wait(), we signal the completion of tasks by probing status files. Because S3 does not provide list-after-write consistency (which might also apply to other storage backend). List() can be only used as an optimization but not a timely way to signal completion. Thus, our strategy is to:
    1) do list()
    2) use get() to signal N tasks that do not show up in 1)
    3) repeat 2) if all N tasks completed, otherwise stop
Note: a small N is probably preferred here. N is set to 4.
#160

### DIFF
--- a/pywren/wait.py
+++ b/pywren/wait.py
@@ -115,7 +115,7 @@ def _wait(fs, THREADPOOL_SIZE):
         callids_done.update(callids_found)
 
         # break if not all N tasks completed
-        if (len(fs_found) < len(fs_samples)):
+        if (len(callids_found) < len(fs_samples)):
             break
         # calculate new still_not_done_futures
         still_not_done_futures = [f for f in not_done_futures if (f.call_id not in callids_done)]
--- a/pywren/wait.py
+++ b/pywren/wait.py
@@ -99,7 +99,7 @@ def _wait(fs, THREADPOOL_SIZE):
     num_samples = 4
     still_not_done_futures = [f for f in not_done_futures if (f.call_id not in callids_done)]
     def fetch_status(f):
-        return storage_handler.get_callset_status(f.callset_id, f.call_id)
+        return storage_handler.get_call_status(f.callset_id, f.call_id)
 
     pool = ThreadPool(num_samples)
     # repeat util all futures are done
--- a/pywren/wait.py
+++ b/pywren/wait.py
@@ -105,7 +105,7 @@ def _wait(fs, THREADPOOL_SIZE):
     # repeat util all futures are done
     while still_not_done_futures:
         fs_samples = random.sample(still_not_done_futures,
-                                  min(num_samples, len(still_not_done_futures)))
+                                   min(num_samples, len(still_not_done_futures)))
         fs_statuses = pool.map(fetch_status, fs_samples)
 
         callids_found = [fs_samples[i].call_id for i in range(len(fs_samples))
--- a/pywren/wait.py
+++ b/pywren/wait.py
@@ -33,11 +33,19 @@ def wait(fs, return_when=ALL_COMPLETED, THREADPOOL_SIZE=64,
     """
     N = len(fs)
 
+
+    # These are performance-related settings that we may eventually
+    # want to expose to end users:
+    MAX_DIRECT_QUERY_N = 16
+    RETURN_EARLY_N = 16
+
     if return_when == ALL_COMPLETED:
         result_count = 0
         while result_count < N:
 
-            fs_dones, fs_notdones = _wait(fs, THREADPOOL_SIZE)
+            fs_dones, fs_notdones = _wait(fs, RETURN_EARLY_N,
+                                          MAX_DIRECT_QUERY_N,
+                                          THREADPOOL_SIZE)
             result_count = len(fs_dones)
 
             if result_count == N:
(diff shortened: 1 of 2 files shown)


pywren/wait.py
```diff
@@ -81,10 +82,47 @@ def _wait(fs, THREADPOOL_SIZE):
...
+        fs_statuses = pool.map(fetch_status, fs_samples)
+
+        callids_found = [fs_samples[i].call_id for i in range(len(fs_samples))
+                         if (fs_statuses[i] is not None)]
+
+        # update done call_ids
+        callids_done.update(callids_found)
+
+        # break if not all N tasks completed
+        if (len(fs_found) < len(fs_samples)):
```
shivaram: why should we break if not all N tasks completed ? (also where is fs_found defined)

ooq: Here is a trade-off. Think this block runs when the tasks are far from finish. Then we are essentially scanning all tasks, and doing this many times before they actually complete.  (yeah, that's a bug. I fixed it.)

ericmjonas:
@ooq Do we have any plots or example tests to show how this mitigates stragglers?

pywren/wait.py
shivaram: Sorry I am still missing something. So if we break out of it during say the first few blocks we sample then how do we get back into this loop for the last few tasks ?

pywren/wait.py
```diff
@@ -81,10 +82,47 @@ def _wait(fs, THREADPOOL_SIZE):
...
     callids_done = set(callids_done)
 
+    num_samples = 4
+    still_not_done_futures = [f for f in not_done_futures if (f.call_id not in callids_done)]
+    def fetch_status(f):
+        return storage_handler.get_call_status(f.callset_id, f.call_id)
+
+    pool = ThreadPool(num_samples)
+    # repeat util all futures are done
+    while still_not_done_futures:
```
ericmjonas: Is this actually the design contract of `_wait`? I thought it would perform _one pass_ of checking if things are done and downloading those; it would not block waiting for all to finish (note the original `_wait` has no looping, that's handeled in `wait()`)

ooq: It's not block waiting for all to finish, with `if (len(callids_found) < len(fs_samples)): break`

ericmjonas: I see so it's still blocking, just not for all of them. Would it make sense to have `num_samples` be passed in (and the policy managed at the layer above) and just have it make a single best-effort to get this number of samples? That is, is there a reason to have this control logic at this layer but the all_completed logic at a layer higher?

shivaram: Yeah it'll be good to have a clean contract between the two functions.

ooq: So say all 100 tasks finish but they only show up with list() 30 seconds after.
With this alternative solution, it would take min(100/N * wait_interval, 30) to find them.

ooq:
@ericmjonas btw, I'll make some plot examples today.

ericmjonas:
@ooq That's awesome, is there a way we could pass the "num_samples" into the outer `wait` in a clean way so that we can have code going forward to show the differences? @shivaram does this seem sane or a proliferation of rarely-used options?

pywren/wait.py
ooq: I actually think that we assume `get_callset_status` and `get_call_status` are instantaneous, then the contract is exactly same here, i.e., use whatever way to give one pass and returns.

ericmjonas:
Ok I spent a bit of time tightening the code up, making the invocation semantics more clear, and making parts of it a little bit more pythonic. I also renamed some variables so it is hopefully now a bit easier to follow along. @ooq  @shivaram would love your opinions

pywren/wait.py
```diff
@@ -79,12 +110,57 @@ def _wait(fs, THREADPOOL_SIZE):
...
+    # we have return_n done.
+    query_count = 0
+    max_queries = min(max_direct_query_n, len(still_not_done_futures))
+
+    if random_query:
+        random.shuffle(still_not_done_futures)
+
+    while query_count < max_queries:
+
+        if len(done_call_ids) >= return_early_n:
```
ooq: I think with this, we assume that there are few stragglers? Let's say we have 100 stragglers (not showing up with list), then it will take  100/return_early_n * WAIT_DUR_SEC to finish. I personally have seen the case with a large number of stragglers yet, but just want to point out the possibility.

pywren/wait.py
```diff
@@ -79,12 +110,57 @@ def _wait(fs, THREADPOOL_SIZE):
...
+            break
+        num_to_query_at_once = THREADPOOL_SIZE
+        fs_to_query = still_not_done_futures[query_count:query_count + num_to_query_at_once]
```
ooq: I think `query_count + num_to_query_at_once` could go out of bound?

ooq:
Thanks for taking it over, @ericmjonas . The code is clean to follow!
It looks good, and I think the corner case pointed out in my comment is not an immediate concern.
I'm fine merging it after the out-of-bound concern is resolved.

pywren/wait.py
ericmjonas: Yeah I think in the future if we want and this becomes a performance pain point we can make return_early_n to be a user-facing parameter.

pywren/wait.py
ericmjonas: Python list slices always return as much as they can:
```
In [3]: x = list(range(10))

In [4]: x
Out[4]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

In [5]: x[5:15]
Out[5]: [5, 6, 7, 8, 9]
```

so I think this should be ok?

Issue #105, opened 2017-03-26
Too big of runtime


ericmjonas:
If a user creates a runtime that unpacks to be too large, make sure we deliver sane errors to the user.

ericmjonas:
In particular the wrenhandler tries to download the pickled objects like data and the function before it deals with the handler. So this is a problem for multiple reasons

- [ ] If func or data is too large, we will error
- [ ] if there is left-over crap from a previous run, we will error

ericmjonas:
A reasonable workaround if the runtime ate up too much space on the last invocation can be to call `map` with `use_cached_runtime=False`

ericmjonas:
Closed by PR #172

PR #172, opened 2017-08-25
Lambda toobig seatbelts

ericmjonas:
This is a series of fixes to handle:
1. if the runtime is too big, communicate to the user and execute proper cleanup on the worker so the worker isn't forever broken. (issue #105)

2. If downloading the func or the data is too large for the remaining space, error as well.

### DIFF
--- a/pywren/wrenhandler.py
+++ b/pywren/wrenhandler.py
@@ -32,6 +32,7 @@ logger = logging.getLogger(__name__)
 
 PROCESS_STDOUT_SLEEP_SECS = 2
 
+# Make sure we have at least 10 MB
 TMP_MIN_FREE_SPACE_BYTES = 10000000
 
 def get_key_size(s3client, bucket, key):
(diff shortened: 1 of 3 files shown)


ericmjonas:
This brings up the question: Should we download the runtime before the user data/func pickles?

pywren/wrenhandler.py
```diff
@@ -32,6 +32,8 @@
 
 PROCESS_STDOUT_SLEEP_SECS = 2
 
+TMP_MIN_FREE_SPACE_BYTES = 10000000
```
shivaram: nit - can we add a comment above this "Leave at least 10MB free space" or something like that

pywren/wrenhandler.py
```diff
@@ -87,8 +96,12 @@ def download_runtime_if_necessary(s3_client, runtime_s3_bucket, runtime_s3_key):
     condatar = tarfile.open(
         mode="r:gz",
         fileobj=wrenutil.WrappedStreamingBody(res['Body'], res['ContentLength']))
-
-    condatar.extractall(runtime_etag_dir)
+    try:
+        condatar.extractall(runtime_etag_dir)
+    except:
```
shivaram: Is there a particular exception that indicates the tarball was too big ? I guess there could be other errors like corrupt runtimes etc.

pywren/wrenhandler.py
```diff
@@ -174,16 +187,26 @@ def generic_handler(event, context_dict):
...
 
-            KS = get_key_size(s3_client, s3_bucket, data_key)
+            data_key_size = get_key_size(s3_client, s3_bucket, data_key)
         if not event['use_cached_runtime']:
             subprocess.check_output("rm -Rf {}/*".format(RUNTIME_LOC), shell=True)
-
-
+        func_key_size = get_key_size(s3_client, s3_bucket, func_key)
+
+        free_disk_bytes = free_disk_space("/tmp")
```
shivaram: is `/tmp` hardcoded elsewhere as well or is there some variable we can use here cc @apengwin

apengwin: `/tmp` is hardcoded in current wrenhandler. The portability PR abstracts this out to a generic `TEMP` variable that's set at the beginning, depending on what OS is running

pywren/wrenhandler.py
ericmjonas: Good question, I will dig further!

pywren/wrenhandler.py
ericmjonas: It is actually hardcoded in the runtimes as well. Unfortunately there are some absolute path dependencies that continue to be a problem in the way conda is packaged.

pywren/wrenhandler.py
apengwin: This could be `1e8` instead?

ericmjonas:
So remember right now we only support running under the python2 runtime for the handler; it's for all the exceptions that I wouldn't otherwise catch, to make sure we have a way to clean up the extracted runtime.

shivaram:
Ah got it. I forgot the handler was python2 specific. Sounds good.

Issue #176, opened 2017-09-02
Create max list seatbelt


ericmjonas:
For the RISE camp we want to make it hard for people to accidentaly laucnh 1e6 jobs
so we should add a config option

shivaram:
Will this be an option from the pywren config file ?

ericmjonas:
Yeah that's the thought. And we'll hand out the rate-limited config files
to the risecamp users. Which means I also need to generate them asap.

PR #177, opened 2017-09-03
New max-limit seatbelt

ericmjonas:
This is to solve #176 for Risecamp @shivaram @ooq  could you guys take a quick look


ooq:
LGTM. Should we make it possible to pass the parameter in map()?

ericmjonas:
@ooq No I want it to be hard for them to override
@shivaram I'll change it to scheduler!

PR #180, opened 2017-09-04
Migrate s3io to jobrunner

ericmjonas:
Right now we save func and data to disk and THEN invoke jobrunner, with the idea of jobrunner being as simple as possible. As a result, the size of func + data + runtime < worker /tmp space. It's not unreasonable to expect data to sometimes be large. We then load func and data into memory anyway. This PR shifts reading those items from S3 into `jobrunner.py`. 

Advantages:
- Now func and data can be as large as memory! 
- Wrenhandler becomes simpler

Disadvantages:
- Slightly harder to get logging information out of jobrunner for when access fails. 
- Harder to get s3 download metrics out as well

I wanted to solicit feedback on this early.  I think this is a much better approach and IIRC @Vaishaal was explicitly asking for it.

### DIFF
--- a/pywren/executor.py
+++ b/pywren/executor.py
@@ -206,13 +206,8 @@ class Executor(object):
                         mod_paths.remove(mod_path)
 
         module_data = create_mod_data(mod_paths)
-        func_str_encoded = func_str
-        #debug_foo = {'func' : func_str_encoded,
-        #             'module_data' : module_data}
-
-        #pickle.dump(debug_foo, open("/tmp/py35.debug.pickle", 'wb'))
         ### Create func and upload
-        func_module_str = pickle.dumps({'func' : func_str_encoded,
+        func_module_str = pickle.dumps({'func' : func_str,
                                         'module_data' : module_data}, -1)
         host_job_meta['func_module_str_len'] = len(func_module_str)
 
--- a/pywren/executor.py
+++ b/pywren/executor.py
@@ -1,11 +1,11 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
-import json
 import logging
 import random
 import time
 from multiprocessing.pool import ThreadPool
+from six.moves import cPickle as pickle
 
 import boto3
 
@@ -206,14 +206,14 @@ class Executor(object):
                         mod_paths.remove(mod_path)
 
         module_data = create_mod_data(mod_paths)
-        func_str_encoded = wrenutil.bytes_to_b64str(func_str)
+        func_str_encoded = func_str
         #debug_foo = {'func' : func_str_encoded,
         #             'module_data' : module_data}
 
         #pickle.dump(debug_foo, open("/tmp/py35.debug.pickle", 'wb'))
         ### Create func and upload
-        func_module_str = json.dumps({'func' : func_str_encoded,
-                                      'module_data' : module_data})
+        func_module_str = pickle.dumps({'func' : func_str_encoded,
+                                        'module_data' : module_data}, -1)
         host_job_meta['func_module_str_len'] = len(func_module_str)
 
         func_upload_time = time.time()
--- a/pywren/jobrunner.py
+++ b/pywren/jobrunner.py
@@ -4,12 +4,20 @@ import base64
 import json
 import sys
 import traceback
+import boto3
+from pywren.wrenutil import WrappedStreamingBody
 
 from six.moves import cPickle as pickle
 from tblib import pickling_support
 
 pickling_support.install()
 
+# def s3_url_parse(url):
+#     if url[:5] != "s3://":
+#         raise Exception("improperly formatted s3 url: {}".format(url))
+#     bucket, key = url[5:].split("/", 1)
+#     return bucket, key
+    
 def b64str_to_bytes(str_data):
     str_ascii = str_data.encode('ascii')
     byte_data = base64.b64decode(str_ascii)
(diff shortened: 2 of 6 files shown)


shivaram:
@ericmjonas Do we want this for 0.3 or can we take some more time to review this

ericmjonas:
We can totally take more time; i don't think it's necessary for 0.3

pywren/wrenhandler.py
```diff
@@ -294,11 +294,21 @@ def generic_handler(event, context_dict):
         CONDA_PYTHON_PATH = "/tmp/condaruntime/bin"
         CONDA_PYTHON_RUNTIME = os.path.join(CONDA_PYTHON_PATH, "python")
 
-        cmdstr = "{} {} {} {} {}".format(CONDA_PYTHON_RUNTIME,
-                                         jobrunner_path,
-                                         func_filename,
-                                         data_filename,
-                                         output_filename)
+        # pass a full json blob
+        jobrunner_config_filename = "/tmp/jobrunner.config.json"
```
shivaram: can we make this a constant at the top of the file ? Also generating a unique name might be good ?

pywren/wrenhandler.py
```diff
@@ -244,33 +244,33 @@ def generic_handler(event, context_dict):
         logger.info("data data download complete, took {:3.2f} sec".format(data_download_time))
         response_status['data_download_time'] = data_download_time
 
-        # now split
-        d = json.load(open(func_filename, 'r'))
+        # d = json.load(open(func_filename, 'r'))
```
shivaram: So I'm not sure I fully follow why we still have the data_key download and func_key download happening in wrenhandler -- isn't the func object the thing we want to move to memory as it gets to 25MB or so from notebooks ?

pywren/jobrunner.py
```diff
@@ -1,33 +1,88 @@
...
+    jobrunner_config_filename = sys.argv[1]
+    out_filename = sys.argv[2]
+
+
     # initial output file in case job fails
     pickle.dump({'result' : None,
                  'success' : False},
                 open(out_filename, 'wb'), -1)
+    jobrunner_config = json.load(open(jobrunner_config_filename,
+                                      'r'))
```
shivaram: really minor nit - i think the `r` can go in the previous line

pywren/jobrunner.py
```diff
@@ -1,33 +1,88 @@
...
+
+
+    # FIXME someday switch to storage handler
+    # download the func data into memory
+    s3_client = boto3.client("s3")
+    func_bucket, func_key = jobrunner_config['func_bucket'], jobrunner_config['func_key']
+    data_bucket, data_key = jobrunner_config['data_bucket'], jobrunner_config['data_key']
+
+    data_byte_range = jobrunner_config['data_byte_range']
+    func_obj_stream = s3_client.get_object(Bucket=func_bucket, Key=func_key)
```
shivaram: this is already supported by the storage API right ? is there any reason we can't directly use it ?

pywren/executor.py
```diff
@@ -194,14 +194,14 @@ def map(self, func, iterdata, extra_env=None, extra_meta=None,
...
-        func_str_encoded = wrenutil.bytes_to_b64str(func_str)
+        func_str_encoded = func_str
         #debug_foo = {'func' : func_str_encoded,
         #             'module_data' : module_data}
 
         #pickle.dump(debug_foo, open("/tmp/py35.debug.pickle", 'wb'))
         ### Create func and upload
-        func_module_str = json.dumps({'func' : func_str_encoded,
-                                      'module_data' : module_data})
+        func_module_str = pickle.dumps({'func' : func_str_encoded,
```
shivaram: is there a reason for switching from json to pickle here ? its a bit weird as we already have the func_str pickled if i'm reading the code right

pywren/jobrunner.py
```diff
@@ -1,33 +1,88 @@
...
+    loaded_func = pickle.loads(loaded_func_all['func'])
+
+    extra_get_args = {}
+    if data_byte_range is not None:
+        range_str = 'bytes={}-{}'.format(*data_byte_range)
+        extra_get_args['Range'] = range_str
+    data_obj_stream = s3_client.get_object(Bucket=data_bucket,
+                                           Key=data_key, **extra_get_args)
+    # FIXME make this streaming
+    loaded_data = pickle.loads(data_obj_stream['Body'].read())
```
shivaram: why is this not streaming right now ? you mean you want to stream the data into the function ?

pywren/wrenhandler.py
```diff
@@ -244,33 +244,33 @@ def generic_handler(event, context_dict):
...
-            full_filename = os.path.join(to_make, os.path.basename(m_filename))
-            #print "creating", full_filename
-            fid = open(full_filename, 'wb')
-            fid.write(b64str_to_bytes(m_data))
-            fid.close()
-        logger.info("Finished writing {} module files".format(len(d['module_data'])))
-        logger.debug(subprocess.check_output("find {}".format(PYTHON_MODULE_PATH), shell=True))
-        logger.debug(subprocess.check_output("find {}".format(os.getcwd()), shell=True))
+
+        # # get modules and save
```
shivaram: again lets delete instead of commenting :)

pywren/executor.py
```diff
@@ -194,14 +194,14 @@ def map(self, func, iterdata, extra_env=None, extra_meta=None,
                         mod_paths.remove(mod_path)
 
         module_data = create_mod_data(mod_paths)
-        func_str_encoded = wrenutil.bytes_to_b64str(func_str)
+        func_str_encoded = func_str
```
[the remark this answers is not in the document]
ericmjonas: yes!

pywren/executor.py
ericmjonas: we first want to unpickle the modules and get them set up in the `PYTHONPATH` before we unpickle the function, otherwise we will get an error unpickling. Thus the two-stage process.

shivaram: The two stage process is fine -- i was just curious why we changed the second layer or serialization from json to pickle. I might be missing something basic ?

pywren/jobrunner.py
ericmjonas: We're calling `.read()` which reads the entire object into memory as a string, and then `loads` which unpickles the string. The "right" thing to do would be pass the `data_object_stream['Body']` to `pickle.load` which wants a file-like object. The problem is that the streaming file-like object boto provides doesn't actually provide _enough_ of the file object API for pickle to use it. We could have a wrapper (like we do for the call to `tarfile`) but even that one is insufficient. 

So for N bytes of input data this means that we briefly have 2N. That said, the temporary string created is easy for GC to get rid of, and this is currently an order of magnitude better than what we had been storing in /tmp/

pywren/wrenhandler.py
ericmjonas: Sorry, again, this is in-progress, there's some detritus that needs to be cleaned out. I should have been more clear, i'm sorry!

pywren/executor.py
ericmjonas: We only were using json because we needed wrenhandler, which runs in the lambda-supplied python (2.7 until recently) to write the `foo.py` module files, and  pickle isn't stable cross-python versions. So we needed something independent there, and then internally to that json object we were storing the base64-encoded pickled function -- because pickling results in a binary stream and we then need to base64 it for json. 

In short, pickle is faster to deserialize, more space-efficient, prevents us from having to b64-encode the internal pickled object, and appropriate because the remote `jobrunner` is running the same version of python as the user.

apengwin:
Why don't we write the `output.pickle` directly from s3 from jobrunner, instead of writing it to /tmp?

ericmjonas:
@apengwin That's a REALLY GOOD QUESTION we should do that. Thank you!

pywren/wrenhandler.py
```diff
@@ -211,67 +204,13 @@ def generic_handler(event, context_dict):
...
-
-        data_download_time = time.time() - start_time
-        logger.info("data data download complete, took {:3.2f} sec".format(data_download_time))
-        response_status['data_download_time'] = data_download_time
-
-        # now split
-        d = json.load(open(func_filename, 'r'))
+        response_status['free_disk_bytes'] = free_disk_bytes
+        # clean up for modules
         shutil.rmtree(PYTHON_MODULE_PATH, True) # delete old modules
```
shivaram: should we also move this inside the jobrunner ? I'm not sure it matters but it'll make the behavior same as before (i.e. download func and if that succeeds then do rmtree)

ericmjonas: yeah good catch will do

PR #192, opened 2017-09-11
add download stats back in, fixes 188

### LINKED ISSUE #188: Add download statistics back for func/data/output
We discarded these when we migrated S3 operations to `jobrunner.py` -- we should put them back in for instrumentation.

ericmjonas:
We removed func/data download metrics in the migration to jobrunner; this restores them by adding them to the jobrunner. Jobrunner now just writes a simple text file that we can read in. It's a text file so that we can write it line-by-line and hopefully still have some stats even if the job dies. 

Fixes #188

### DIFF
--- a/pywren/jobrunner.py
+++ b/pywren/jobrunner.py
@@ -4,7 +4,9 @@ import base64
 import shutil
 import json
 import sys
+import time
 import boto3
+
 
 from six.moves import cPickle as pickle
 from tblib import pickling_support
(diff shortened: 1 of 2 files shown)


pywren/jobrunner.py
```diff
@@ -41,9 +43,21 @@ def b64str_to_bytes(str_data):
 output_bucket = jobrunner_config['output_bucket']
 output_key = jobrunner_config['output_key']
 
+## Jobrunner stats are fieldname float
+jobrunner_stats_filename = jobrunner_config['stats_filename']
+# open the stats filename
+stats_fid = open(jobrunner_stats_filename, 'w')
+
+def write_stat(stat, val):
+    stats_fid.write("{} {:f}\n".format(stat, val))
```
shivaram: can we also `flush` the file handle here ? it'll make sure full lines are written to the file system asap

ericmjonas:
Good catch on the flush!

Comparison is tricky -- I don't know if anyone ever actually looked at the
old stats; right now I'm adding them back in because I want to build a
complete timeline of jobs for group meeting.

Issue #186, opened 2017-09-08
restore __version__


ericmjonas:
Restore user-accessable version information `pywren.__version__`

PR #193, opened 2017-09-11
Fix issue 186

ericmjonas:
Fixes issue #186 restoring `__version__` but also fixes some test case bugs:
1. IAM instance-profile information takes a bit to propagate so standalone tests sometimes fail
2. the dummy-invoker tests sometimes is a bit slow due to downloading runtime ; we extended its max runtime to compensate.

### DIFF
--- a/tests/ec2_standalone_tests.sh
+++ b/tests/ec2_standalone_tests.sh
@@ -4,7 +4,18 @@ set -x
 if [ "$RUN_STANDALONE" != "true" ]; then
     exit 0
 fi
+# sometimes the instance profile isn't visible yet to the function.
+# this results in erratic test behavior and tests sometimes failing
 
+n=0
+until [ $n -ge 5 ]
+do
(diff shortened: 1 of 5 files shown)


tests/ec2_standalone_tests.sh
```diff
@@ -4,7 +4,18 @@ set -x
 if [ "$RUN_STANDALONE" != "true" ]; then
     exit 0
 fi
+# sometimes the instance profile isn't visible yet to the function.
+# this results in erratic test behavior and tests sometimes failing
 
+n=0
+until [ $n -ge 5 ]
+do
+    aws iam get-instance-profile --instance-profile-name pywren_travis_$BUILD_GUID && break  # substitute your command here
```
shivaram: do you want to pipe this output somewhere (i have no idea what `get-instance-profile` returns). Also this comment can be deleted ?

ericmjonas: It returns a json blob of the instance profile, but we already know that; that said it's verbose and I will direct it to `/dev/null`

PR #194, opened 2017-09-11
Created per-handler-type env vars to fix #137

### LINKED ISSUE #137: lapack/blas calls very slow on multicore machines for stand alone mode
Consider the following python script
```
import scipy.linalg
import numpy as np
import time
x = np.random.randn(4096, 4096)
x = x.dot(x.T)
y = np.random.randn(4096, 1000)
t = time.time()
scipy.linalg.solve(x,y, sym_pos=True)
e = time.time()
print("Took {0}".format(e - t))`
```

The default runtimes take > 135 seconds to execute the above code. Whereas a standard anaconda install takes ~2 seconds to execute the above code. This was observed on a c3.8xlarge which has 16 hardware cores.

ericmjonas:
Fixes the problem that Viashaal noticed with OMP_NUM_THREADS issue #137

### DIFF
--- a/pywren/wrenhandler.py
+++ b/pywren/wrenhandler.py
@@ -136,7 +136,10 @@ def aws_lambda_handler(event, context):
         'log_group_name' : context.log_group_name,
         'log_stream_name' : context.log_stream_name,
     }
-    return generic_handler(event, context_dict)
+    # MKL does not currently play nicely with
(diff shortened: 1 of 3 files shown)


ericmjonas:
Not that guy at all, I sincerely believed we already had them! We need them and the logic is easy.

Issue #187, opened 2017-09-09
Create ability to cancel jobs


ericmjonas:
Only slightly useful for lambdas, but really useful for standalone.

PR #196, opened 2017-09-12
Add ability to cancel a job [WIP]

ericmjonas:
# What 
This adds the ability to cancel a job, by making `.cancel()` method of a future actually do something.

# Why
- If you launch 1e4 lambdas and then think "oops!" this lets you bleed through the queue quickly -- the lambdas will each run for like 100ms to clear out the queue. 
- if you are using standalone mode for long-running tasks like deepnets and want to abort the remote work, this is the only supported way of saying "oops" and actually cancelling the deepnet process while it is running. 

# How 
 All it does is write a `cancel` key to s3; which our runner periodically polls for. We check for the cancel job id

1. at the very beginning of the job
2. every `CANCEL_CHECK_EVERY_SECS` while running the actual function 

and then return an exception for the canceled functions. 

Note that a function's "canceled" status isn't set on the client side but rather the client side writes the key and then relies on the remote to signal (via the exception) that it was canceled. Sometimes functions may be canceled but it's noticed remotely too late, or they are already done. 

*Looking for feedback here. *

This once again exposed some of the weaknesses in our storage configuration abstraction; once again the future has to create a default storage handler inside itself. 

associated with issue #187

### DIFF
--- a/pywren/future.py
+++ b/pywren/future.py
@@ -58,6 +58,13 @@ class ResponseFuture(object):
         self._state = new_state
 
     def cancel(self):
+        storage_config = wrenconfig.extract_storage_config(wrenconfig.default())
+        storage_handler = storage.Storage(storage_config)
+
+        call_status = storage_handler.get_call_status(self.callset_id, 
+                                                      self.call_id)
+
+
         raise NotImplementedError("Cannot cancel dispatched jobs")
 
     def cancelled(self):
--- a/pywren/future.py
+++ b/pywren/future.py
@@ -58,14 +58,13 @@ class ResponseFuture(object):
         self._state = new_state
 
     def cancel(self):
+        # FIRST This is annoying; how will it have the custom access
+        # to the storage handler information
+
         storage_config = wrenconfig.extract_storage_config(wrenconfig.default())
         storage_handler = storage.Storage(storage_config)
-
-        call_status = storage_handler.get_call_status(self.callset_id, 
-                                                      self.call_id)
-
-
-        raise NotImplementedError("Cannot cancel dispatched jobs")
+        storage_handler.put_canceled(self.callset_id,
+                                     self.call_id, "foo")
 
     def cancelled(self):
         raise NotImplementedError("Cannot cancel dispatched jobs")
@@ -80,7 +79,8 @@ class ResponseFuture(object):
             return False
         return True
 
-    def result(self, timeout=None, check_only=False, throw_except=True, storage_handler=None):
+    def result(self, timeout=None, check_only=False,
+               throw_except=True, storage_handler=None):
         """
 
 
@@ -154,6 +154,10 @@ class ResponseFuture(object):
             elif exception_args[0] == "OUTATIME":
                 if throw_except:
                     raise Exception("process ran out of time")
+                return None
+            elif exception_args[0] == "CANCELED":
+                if throw_except:
+                    raise Exception("job was canceled")
                 return None
             else:
                 if throw_except:
--- a/pywren/storage/storage.py
+++ b/pywren/storage/storage.py
@@ -5,7 +5,8 @@ import os
 
 from  .exceptions import StorageNoSuchKeyError, StorageOutputNotFoundError
 from .s3_backend import S3Backend
-from .storage_utils import create_status_key, create_output_key, status_key_suffix, create_cancel_key
+from .storage_utils import create_status_key, create_output_key
+from .storage_utils import status_key_suffix, create_cancel_key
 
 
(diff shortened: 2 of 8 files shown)


zehric:
The reason this was failing previously had to do with the difference in spelling of "cancelled" and "canceled" (British and American English). I went with "cancelled" because that seems to be what Python uses.

Another note is that if this is to be merged in with #281, there will have to be a bit more refactoring to make the wrenhandler changes work on Windows as well.

pywren/wrenhandler.py
```diff
@@ -59,6 +59,7 @@
 logger = logging.getLogger(__name__)
 
 PROCESS_STDOUT_SLEEP_SECS = 2
+CANCEL_CHECK_EVERY_SECS = 5.0
```
Vaishaal: Does this have to be hard-coded to 5, can we make this configurable?

pywren/future.py
```diff
@@ -73,8 +73,16 @@ def _set_state(self, new_state):
         ## FIXME add state machine
         self._state = new_state
 
-    def cancel(self):
-        raise NotImplementedError("Cannot cancel dispatched jobs")
+    def cancel(self, storage_handler=None):
+        # FIRST This is annoying; how will it have the custom access
```
shivaram: Can we rewrite this comment ? I dont understand what this says

pywren/future.py
```diff
@@ -73,8 +73,16 @@ def _set_state(self, new_state):
...
+        # to the storage handler information
+
+        if storage_handler is None:
+            storage_config = wrenconfig.extract_storage_config(wrenconfig.default())
+            storage_handler = storage.Storage(storage_config)
+
+        storage_handler.put_cancelled(self.callset_id,
+                                      self.call_id, "foo")
```
shivaram: Can we not hard code "foo" in the code ? Create a constant ? Also will this field be used for anything else in the future ? If so can we create a map that says "STATUS_CANCELED: 1" and write the value from that map here ?

pywren/storage/storage.py
```diff
@@ -67,6 +68,17 @@ def put_func(self, key, func):
...
         return self.backend_handler.put_object(key, func)
 
+    def put_cancelled(self, callset_id, call_id, cancelled_data):
+        """
+        Put data to signal cancel into storage
+        :param key: cancel_key
+        :param func: cancelled_data
+        :return: None
+        """
+        key = create_cancel_key(self.prefix, callset_id,
```
shivaram: why do we call create_cancel_key again here ? isn't create_keys called before put_cancelled is ?

pywren/wrenhandler.py
```diff
@@ -225,6 +229,13 @@ def generic_handler(event, context_dict, custom_handler_env=None):
         status_key = event['status_key']
         func_key = event['func_key']
         data_key = event['data_key']
+        cancel_key = event['cancel_key']
+
+        # Check for cancel
+        if key_exists(s3_client, s3_bucket, cancel_key):
+            raise Exception("Function cancelled")
```
shivaram: shouldn't this be `Exception("CANCELLED", "Function cancelled")` ?

shivaram: Similar to below can we add a logger.info saying invocation was canceled before start or something like that ?

pywren/storage/storage.py
zehric: I think you are right about this. Will remove in the next update.

pywren/wrenhandler.py
zehric: You're correct, will fix this

zehric: That sounds good, will add it in

pywren/wrenhandler.py
zehric: I think we can leave it hardcoded at 5 for now. It shouldn't be difficult to change and I don't see much benefit of allowing the user to specify it.

pywren/future.py
zehric: Since the only information we need to cancel a job is the _existence_ of a key, I don't think the value is meaningful. We'd never need to change it. In this case, I don't think hardcoding a value is bad; I feel that creating such a map would create unnecessary clutter. I will change the constant from "foo" to "CANCEL", however.

pywren/storage/storage.py
zehric: Actually, I think the reason behind this is that the cancel key isn't accessible from a `future`, whereas the `call_id` and `callset_id` are. Since `create_cancel_key` is just a path join of these ids. I will leave this in here.

Issue #197, opened 2017-09-13
standalone non-deterministically drops jobs when > 50 instances are spun up


Vaishaal:
https://github.com/pywren/pywren/blob/master/pywren/scripts/standalone.py#L229-L243

Throws an exception when there are a lot of instances (from my experiments > 50 suffices to trigger the bug), because the default EC2 Log Group creation limit is super low (AFAIK almost all the default cloudwatch limits are very low). Since this happens outside a try: except, the manager process just assumes everything was handled successfully and deletes the message from the SQS queue, leading to much frustration.

We could jitter in the job handler, but I'm worried that will make our per job latency even higher since now this has to happen for every job...

Currently I'm commenting this logging code out because personally these logs have never helped me. But what should we do about this issue?

ericmjonas:
Sorry is this log stream or log group limits ? And why would jittering the
log stream creation hurt per job latency ? Log streams are per machien not
per job, right ?

ericmjonas:
@Vaishaal can you get on the eric-fix-standalone branch and see if you still have this problem? It switches from processes to threads and thus only creates one instance of the CloudWatchLogHandler; you might still need to jitter the start-up a bit to prevent the service from being hammered, but in this case explicitly it shouldn't be a problem.

ericmjonas:
I believe this is fixed in branch `eric-fix-standalone`

Issue #185, opened 2017-09-07
Standalone mode tests fail on merge test


ericmjonas:
Github/Travis check two versions of the source code when a pull request is created: the code at the head of the PR and the code that results from an automatic merge of the two codebases. 

Currently passing a sha1 into the launch standalone command can error out on the "post-merge" code base because it probably doesn't exist in the remote repo. Therefor the standalone instance actually _only_ runs pywren off the master branch. This is problematic. You can see the failed attempt to check out the branch via sha1 in the cloudinit.output logs on cloudwatch. 

The better solution is to launch the standalone instance with a pointer to a pywren source install or something like that, or just have it introspect whatever it thinks it needs from the local pywren install.

ericmjonas:
Related: #123

Issue #139, opened 2017-05-16
Standalone Executors without "instance name" errors on launch of pywren server


Vaishaal:
Line 292 - 296 in standalone.py

```
    instance = get_my_ec2_instance(aws_region)
    ec2_metadata = get_my_ec2_meta(instance)
    server_name = ec2_metadata['Name']
    log_format_str ='{} %(asctime)s - %(name)s - %(levelname)s - %(message)s'.format(server_name)
    log_stream_prefix = ec2_metadata['instance_id']

```

if `ec2_metadata['Name'] ` is not defined (because the instances were launched by a script that failed before assigning the name attribute), then the pywren-server will crash and relaunch itself and crash again. To compound to this issue if this happens on a large cluster (lets say 500 instances), then ec2-describe-instances gets rate-limited which means you can't assign a name to any of the instances even after the fact.

Issue #135, opened 2017-05-16
Standalone mode launcher errors when more than 200 instances are launched


Vaishaal:
1) Rate limited by the ec2-describe-instances command
2) Boto3 "filter" function does not support more than 200 instances at once

PR #198, opened 2017-09-13
Standalone bug-fixes

### LINKED ISSUE #189: Executor won't run from shard if numshards = 1

ericmjonas:
First, I want to open with this pull request is too large and is the result of developing off-master for too long. @Vaishaal and I were the main users of standalone and started doing our own thing. Lately, the drift has hurt us. Additionally, now there's too much publicly-pushed code to do a clean rebase.

In short, don't do what I did. 

On to the PR! 
- Fix logging both stdout and stderr on the cloudwatch agent -- before stderr wasn't being sent to cloudwatch
- stop doing an apt-get update/upgrade on the standalone instance when we boot it, as some images have very specific pinned kernel versions for CUDA stuff
- add spot instance support; a lot of this is taken from the flintrock support that I originally wrote; vaishaal and I have both used this very successfully
- Run when there is only a single shard. fixes #189 
- switch standalone runner from using python multiprocessing to python threads. This is actually 
far superior as 
    - closer to the actual lambda model (wrenhandler is invoked in-process)
    - it signfiicantly simplifies the log handling code with Watchtower, preventing the behavior @Vaishaal  was seeing in #197 
- migrate to properly processing the messages with SQS visibility timeout

Merging this is a pre-req for dealing with issues #185 and #153 and #139 and #135


Vaishaal:
Can we add the bootup jitter to this PR?

ericmjonas:
Do we have any reason to think 120s is a good number? That's a lot of jitter!

Vaishaal:
Given that you are launching 200+ instances when this matters, it didn't seem like a lot to me. We could probably make it 60s?

Vaishaal:
Other than that the rest of the changes look good to me!

Vaishaal:
LGTM

Issue #199, opened 2017-09-14
Currently no way to check if future success or failure


ericmjonas:
There is no convenient way to check if a future's current state is success or failure. 

We should add `.success()` and `.failure()` methods

PR #200, opened 2017-09-15
Fix issue 199

ericmjonas:
Fix issue #199 and additional make `.done()` behave correctly (how did we never have a test for this?!)


pywren/future.py
```diff
@@ -94,6 +102,10 @@ def result(self, timeout=None, check_only=False, throw_except=True, storage_hand
         if self._state == JobState.new:
             raise ValueError("job not yet invoked")
 
+        if check_only:
```
shivaram: why don't we write a new `check` function that we can call from within this function ? That way we can avoid that `FIXME` below ?

pywren/future.py
```diff
@@ -172,36 +187,34 @@ def result(self, timeout=None, check_only=False, throw_except=True, storage_hand
...
+            if throw_except:
 
-        elif throw_except:
-
-            self._exception = call_invoker_result['result']
-            self._traceback = (call_invoker_result['exc_type'],
-                               call_invoker_result['exc_value'],
-                               call_invoker_result['exc_traceback'])
-
-            self._state = JobState.error
```
shivaram: We seem to have dropped this line `self._state = JobState.error` -- is this intentional ?

ericmjonas: we switched to using the proper `self._set-state(JobState.error)` in line 195

pywren/future.py
ericmjonas: Yes that's the right thing to do but the three of us all have deadlines; writing `check()` creates _another_ "temporarily create the storage handler` point in the code and I don't want to think about that right now.  

basically our semantics are crap and we need a refactoring; right now I just want `done()` to work well enough that I can have a fast status check. 

Basically this is meant as a patch not a refactor :) 

I agree though this all stinks, it stinks to high heaven.

PR #210, opened 2017-10-04
[Issue#209] support launch lambda with lambda.

ooq:


pywren/wrenhandler.py
```diff
@@ -129,8 +130,24 @@ def b64str_to_bytes(str_data):
     return byte_data
 
 
+def invoke_tasks(function_name, tasks):
+    logger.info("invoking tasks")
+    lambda_client = boto3.client('lambda')
+    pool = ThreadPool(min(64, len(tasks)))
```
apengwin: Is there a reason we use max 64 threads? Just curious

ooq: There is minimal improvement over 64 threads due to inefficiency of boto3. (this is why we want to launch lambda via lambda in the first place)

Issue #211, opened 2017-10-05
half-implemented wrapped stream handler causes runtime untar errors on py2.7


ericmjonas:
Python 2.7's `tarfile` module actually uses `seek` and as a result the `WrappedStreamHandler` we provide is insufficient. This leads to some tarballs of runtimes failing to untar. We should instead
simply use `.read()` with a `BytesIO` object; the memory overhead will be minimal.

PR #212, opened 2017-10-05
move from using wrappedstreamhandler to bytesio and read

ericmjonas:
fixes issue #211

### DIFF
--- a/pywren/wrenhandler.py
+++ b/pywren/wrenhandler.py
@@ -10,6 +10,7 @@ import tarfile
 import time
 import traceback
 from threading import Thread
+import io
 
 import boto3
 import botocore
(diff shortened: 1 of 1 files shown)


pywren/wrenhandler.py
```diff
@@ -92,14 +93,13 @@ def download_runtime_if_necessary(s3_client, runtime_s3_bucket, runtime_s3_key):
...
-
-        condatar = tarfile.open(
-            mode="r:gz",
-            fileobj=wrenutil.WrappedStreamingBody(res['Body'],
-                                                  res['ContentLength']))
+        res_buffer_io = io.BytesIO(res_buffer)
+        condatar = tarfile.open(mode="r:gz", fileobj=res_buffer_io)
         condatar.extractall(runtime_etag_dir)
+        del res_buffer_io # attempt to clear this proactively
+        del res_buffer
```
ooq: I think `del` does not immediately clear up memory. We can enforce gc by calling system.gc()

PR #213, opened 2017-10-10
Travis

### LINKED ISSUE #149: Passing pylint should be a prerequisite for actual travis tests

apengwin:
Fix #149

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ install:
   - tests/install_pywren.sh
 
 before_install:
+  - pip install pylint
   - tests/run_pylint.sh
   
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
   - tests/install_pywren.sh
 
 before_install:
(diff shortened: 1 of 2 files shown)


apengwin:
@shivaram  @ooq  i think this works. I modified `future.py` to test that it fails lint.

shivaram:
Thats great @apengwin - What was the major change compared to the previous PR ?

apengwin:
@shivaram  the problem w the previous approach was that we had a bunch of errors w import botocore and import boto3. This just runs lint after we've set up conda w all the packages.

Issue #220, opened 2017-10-28
[Solution Included]pywren-config default is not properly setuped: making NoSuchBucket error.


Beomi:
## Environ

- OS X 10.12.6 ( Sierra )
- Python3 (aws-cli installed/setuped)

## Situation

When I followed [getting started](http://pywren.io/pages/gettingstarted.html), I setup aws correctly however it doesn't work, providing "NoSuchBucket" Error.

## Error Trace log

```
$ pywren-setup
This is the PyWren interactive setup script
Your AWS configuration appears to be set up, and your account ID is xxxxxxxxxxxx
This interactive script will set up your initial PyWren configuration.
If this is the first time you are using PyWren then accepting the defaults should be fine.
What is your default aws region? [us-west-2]: ap-northeast-2
Location for config file:  [/Users/beomi/.pywren_config]:
PyWren requires an s3 bucket to store intermediate data. What s3 bucket would you like to use? [beomi-pywren-xxx]:
Bucket does not currently exist, would you like to create it? [Y/n]: y
PyWren prefixes every object it puts in S3 with a particular prefix.
PyWren s3 prefix:  [pywren.jobs]: lamlam
Would you like to configure advanced PyWren properties? [y/N]: n
PyWren standalone mode uses dedicated AWS instances to run PyWren tasks. This is more flexible, but more expensive with fewer simultaneous workers.
Would you like to enable PyWren standalone mode? [y/N]: n
Creating config /Users/beomi/.pywren_config
new default file created in /Users/beomi/.pywren_config
lambda role is pywren_exec_role_1
Creating bucket beomi-pywren-xxx.
Creating role.
Deploying lambda.
Pausing for 5 seconds for changes to propagate.
Pausing for 5 seconds for changes to propagate.
Pausing for 5 seconds for changes to propagate.
Successfully created function.
Pausing for 10 sec for changes to propagate.
Traceback (most recent call last):
  File "/Users/beomi/.virtualenvs/tmp/bin/pywren-setup", line 11, in <module>
    sys.exit(interactive_setup())
  File "/Users/beomi/.virtualenvs/tmp/lib/python3.5/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/beomi/.virtualenvs/tmp/lib/python3.5/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
[...]

PR #221, opened 2017-10-28
Make runtime s3_bucket to hard-coded value

Beomi:
Temporary solution for #220 issue.

Hard-code runtime s3 region to `us-west-2`.

This make #220 fixed for now.


Beomi:
Fixed in #220, Close PR.

PR #223, opened 2017-10-29
Process failure properly captured

ericmjonas:
When the function or process dies unexpectedly, we don't currently capture that; it can also cause race conditions with the way we were reading standard out. This PR includes a fix for the race condition, a test for process failure, and a custom error return.


pywren/wrenhandler.py
```diff
@@ -318,9 +324,16 @@ def consume_stdout(stdout, queue):
                     response_status[key] = float_value
 
         end_time = time.time()
+        response_status['retcode'] = process.returncode
```
shivaram: This was already set in line 316 ?

pywren/wrenhandler.py
```diff
@@ -318,9 +324,16 @@ def consume_stdout(stdout, queue):
...
 
         response_status['stdout'] = stdout.decode("ascii")
 
+        if process.returncode != 0:
```
shivaram: Is it better to do this right after the process finishes (line 317) ? I'm not sure if we want to run through the `JOBRUNNER_STATS_FILENAME` loop or not in this case

pywren/wrenhandler.py
```diff
@@ -292,13 +292,18 @@ def consume_stdout(stdout, queue):
...
+        while t.isAlive() or process.returncode is None:
+            logger.info("Running {} {}".format(time.time(), process.returncode))
             try:
                 line = q.get_nowait()
                 stdout += line
                 logger.info(line)
             except Empty:
                 time.sleep(PROCESS_STDOUT_SLEEP_SECS)
+            process.poll() # this updates retcode but does not block
+            if not t.isAlive() and process.returncode is None:
```
shivaram: I'm not sure I follow this -- if the stdout reader thread is not alive then how can we get output from it ? will the `q.get_nowait()` ever give anything useful in that case ?

Ideally I'd think we want `t.isAlive` and `process.retcode is None` in line 295  ?

pywren/wrenhandler.py
ericmjonas: Oops that's embarrassing, will fix

pywren/wrenhandler.py
ooq: For this comment, `q.get_nowait()` should fire `except Empty`, is that correct? @Vaishaal

pywren/wrenhandler.py
```diff
@@ -334,13 +334,18 @@ def consume_stdout(stdout, queue):
         t.start()
 
         stdout = b""
-        while t.isAlive():
+        while t.isAlive() or process.returncode is None:
```
ooq: Are there cases `t.isAlive()` is true, and `process.returncode is None` is false?

ooq:
@Vaishaal I left two comments, but otherwise good. Thanks for picking up this old issue!

Vaishaal:
@ooq can you approve? Travis Seems to pass now.

PR #224, opened 2017-10-31
Standalone multiple executors

Vaishaal:
This is kind of a big change so starting this discussion early so I can make necessary changes. I'll add tests and stuff after you guys agree on the design. 

tl;dr allow for multiple executors on a single standalone box

Basically use the fcntl syscall to acquire a filesystem lock for runtime_downloading/cache_checking. Because of @ericmjonas jobrunner stuff a lot of the annoying files in /tmp are no longer an issue. The lock does mean only one runtime per *machine* so I don't know what we want to do about that? That is if you run 2 pywren jobs right after another with different runtimes there is a potential race condition. I don't want to have a separate runtime for each process because these runtimes get big and we actually hardcode the locations and blah blah. 

Other thing is this will increase our API hammering at boot, so I made that part asynchronous with exponential backoff. 

I know its a little rough but hopefully a couple round trips will make this mergeable.


ericmjonas:
Just to make sure I understand, the current configuration means that each worker can only ever have a single runtime on it; the lock only prevents a race condition while it is being downloaded. Is that correct?

Vaishaal:
yes.

pywren/scripts/standalone.py
```diff
@@ -310,41 +310,57 @@ def server(aws_region, max_run_time, run_dir, sqs_queue_name, max_idle_time,
...
-    logger.addHandler(handler)
-    logger.setLevel(logging.DEBUG)
-
-    wren_log = pywren.wrenhandler.logger
-    wren_log.addHandler(handler)
-
-    wren_log.addHandler(debug_stream_handler)
+    def async_log_setup():
+        ''' None of this stuff should be on the critical path to launching an
+            * instance. Instances should start dequeuing from SQS queue as soon
```
shivaram: So what happens to log messages generated before this comes up ? Will we lose them ? Is that okay to do ? (cc @ericmjonas)

pywren/scripts/standalone.py
```diff
@@ -310,41 +310,57 @@ def server(aws_region, max_run_time, run_dir, sqs_queue_name, max_idle_time,
...
+                handler.setFormatter(formatter)
+                logger.addHandler(handler)
+                logger.setLevel(logging.DEBUG)
+                wren_log = pywren.wrenhandler.logger
+                wren_log.addHandler(handler)
+                wren_log.addHandler(debug_stream_handler)
+                success = True
+            except:
+                backoff_time *= 2
+
```
shivaram: So one thing that will be very good to do is demarcate what will be called once and what will be called once per-job dequeued from SQS. My understanding is that server is called once per machine -- is that right ?

pywren/scripts/standalone.py
```diff
@@ -310,41 +310,57 @@ def server(aws_region, max_run_time, run_dir, sqs_queue_name, max_idle_time,
...
+                debug_stream_handler = logging.StreamHandler()
+                handler.setFormatter(formatter)
+                logger.addHandler(handler)
+                logger.setLevel(logging.DEBUG)
+                wren_log = pywren.wrenhandler.logger
+                wren_log.addHandler(handler)
+                wren_log.addHandler(debug_stream_handler)
+                success = True
+            except:
+                backoff_time *= 2
```
shivaram: Can you add a log message here noting the exception and how long you are going to sleep ?

pywren/wrenhandler.py
```diff
@@ -120,6 +123,7 @@ def download_runtime_if_necessary(s3_client, runtime_s3_bucket, runtime_s3_key):
 
     # final operation
     os.symlink(expected_target, CONDA_RUNTIME_DIR)
+    fcntl.lockf(lock, fcntl.LOCK_UN)
```
shivaram: Can you close the `lock` file here ?

pywren/wrenhandler.py
```diff
@@ -60,6 +61,8 @@ def download_runtime_if_necessary(s3_client, runtime_s3_bucket, runtime_s3_key):
 
     """
 
+    lock = open("/tmp/runtime_download_lock", "a")
+    fcntl.lockf(lock, fcntl.LOCK_EX)
     # get runtime etag
     runtime_meta = s3_client.head_object(Bucket=runtime_s3_bucket,
```
shivaram: I agree that supporting different runtimes can be out of scope for now. But can we detect that a user is trying to do that and fail here ? i.e. if you try to download a new runtime while another handler is using an existing runtime or something like that. Or is there a simpler way to detect this by looking at the job definitions ?

pywren/scripts/standalone.py
Vaishaal: once per "process"

pywren/wrenhandler.py
Vaishaal: I think theres a discussion about this on slack, I think we can support multiple runtimes pretty easily...

pywren/ec2_standalone_files/supervisord.conf
```diff
@@ -48,3 +48,5 @@ files = /etc/supervisord.d/*.conf
 [program:pywren]
 command=/home/ec2-user/anaconda/bin/pywren-server --run_dir {run_dir} --aws_region {aws_region} --sqs_queue_name {sqs_queue_name} --max_idle_time={max_idle_time} --idle_terminate_granularity={idle_terminate_granularity} 
 autorestart=unexpected
+process_name=%(program_name)s-%(process_num)s
```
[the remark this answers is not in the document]
Vaishaal: supervisord provides it

pywren/scripts/standalone.py
Vaishaal: Yes. I think its fine because the alternative is that the executor crashes or hangs until it can set up the logger...

pywren/wrenhandler.py
```diff
@@ -60,6 +61,8 @@ def download_runtime_if_necessary(s3_client, runtime_s3_bucket, runtime_s3_key):
 
     """
 
+    lock = open("/tmp/runtime_download_lock", "a")
```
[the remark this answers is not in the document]
Vaishaal: how do we generate the uuid?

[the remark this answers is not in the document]
Vaishaal: i also don't think its safe to remove it....

pywren/scripts/standalone.py
shivaram: So what happens to log messages generated while the stream is not ready yet. Do they just get dropped ? And is it safe to set this in parallel ? 

It seems to me this is orthogonal to the multi-executor issue as such. Is that right or does this somehow only showup in this case ?

pywren/scripts/standalone.py
```diff
@@ -310,41 +310,57 @@ def server(aws_region, max_run_time, run_dir, sqs_queue_name, max_idle_time,
...
+            * If there are any exceptions in this function,
+            * we should exponentially backoff and try again until we succeed,
+            * this is critical because if this doesn't happen we end up
+            * clogging all EC2 resources
+        '''
+        success = False
+        backoff_time = 5
+        while (not success):
+            try:
+                time.sleep(backoff_time)
```
shivaram: Its usually better to sleep in the `except` block -- that way you are not changing anything for the non-exception cases ?

pywren/scripts/standalone.py
shivaram: No response to this ?

pywren/wrenhandler.py
```diff
@@ -25,11 +26,12 @@
     import wrenutil # pylint: disable=relative-import
     import version  # pylint: disable=relative-import
```
shivaram: Add a comment before this to say that `{0}` will get populated by ETag

pywren/wrenhandler.py
```diff
@@ -236,31 +252,36 @@ def generic_handler(event, context_dict, custom_handler_env=None):
         callset_id = event['callset_id']
         response_status['call_id'] = call_id
         response_status['callset_id'] = callset_id
-
-        CONDA_PYTHON_PATH = "/tmp/condaruntime/bin"
-        CONDA_PYTHON_RUNTIME = os.path.join(CONDA_PYTHON_PATH, "python")
+        runtime_meta = s3_client.head_object(Bucket=runtime_s3_bucket,
+                                             Key=runtime_s3_key)
+        ETag = str(runtime_meta['ETag'])[1:-1]
+        conda_python_path = "/tmp/condaruntime_{0}/bin".format(ETag)
```
shivaram: Use the constant defined above ? `CONDA_RUNTIME_DIR` that is

pywren/wrenhandler.py
```diff
@@ -236,31 +252,36 @@ def generic_handler(event, context_dict, custom_handler_env=None):
...
+        jobrunner_stats_filename = JOBRUNNER_STATS_FILENAME.format(pid)
+        python_module_path = PYTHON_MODULE_PATH.format(pid)
 
         jobrunner_config = {'func_bucket' : s3_bucket,
                             'func_key' : func_key,
                             'data_bucket' : s3_bucket,
                             'data_key' : data_key,
                             'data_byte_range' : data_byte_range,
-                            'python_module_path' : PYTHON_MODULE_PATH,
+                            'python_module_path' : python_module_path.format(pid),
```
shivaram: we have already called `.format(pid)` to get this ?

pywren/scripts/standalone.py
Vaishaal: Good idea. The reason this becomes a problem is because the number of effective executors increases by like 20x after so you'll have much more conflicts with the logging. I figured letting this sort out on a separate thread rather than delaying the execution until all the logging stuff is sorted out.

Vaishaal:
uh by design this test will fail:
https://travis-ci.org/pywren/pywren/jobs/313716721#L1933

(when I change the runtime dir to something other than /tm/condaruntime some of the binaries stop working (seems like conda breaks)) 

What can I do to fix this?

shivaram:
@Vaishaal You can add the new location to `$PATH` ?

Vaishaal:
Not sure if that would fix it, the problem is the shebang in the conda bashscript. You'd have to do some ugly monkey patching to fix.

Vaishaal:
I guess I could create a symlink to /tmp/condaruntime

shivaram:
Creating the symlink wouldn't work if multiple executors run on the same machine ? I guess the broader question is do we really want `conda info` to work from within the lambda ?

Vaishaal:
I would only create the symlink so this test passes. All the executors will still use the other symlink with the custom etag.

shivaram:
Thats not that healthy imho :) The test should be reflective of what users will see. @ericmjonas Do you have thoughts on why we included that test ?

Vaishaal:
Yeah I know. Ok I'll wait on @ericmjonas . I think thats the only test thats failing. I'll try and write a test for the multiple executor version and fix your comments.

pywren/scripts/standalone.py
ericmjonas: Yeah I think it would be worthwile to log this too @Vaishaal

pywren/scripts/standalone.py
ericmjonas: Can you name this constant something like "EXP_BACKOFF_FACTOR" or something? I just hate having these mystery constants hanging out

ericmjonas:
@Vaishaal @shivaram We had wanted for invocations of binaries installed in the runtime which hard-coded the location of the python interpreter (some installed scripts do this, especially `conda`, as it wants to use the "correct" python regardless of what the status is of the `$PATH` is). 

I don't remember why we decided this was a thing we wanted, though -- I'm not sure what the case for calling `conda` from inside a remote worker is? I think the concern was that there might be other utils that had this problem, but I'm willing to disable the test, with a note saying that @Vaishaal's change made us remove this and we will reevaluate if someone cares.

Vaishaal:
Ok thanks eric!

pywren/scripts/standalone.py
Vaishaal: ok handled.

pywren/scripts/standalone.py
Vaishaal: Yeah some messages get dropped. I think its a fine design decision?

pywren/wrenhandler.py
```diff
@@ -236,31 +252,36 @@ def generic_handler(event, context_dict, custom_handler_env=None):
         callset_id = event['callset_id']
         response_status['call_id'] = call_id
         response_status['callset_id'] = callset_id
-
-        CONDA_PYTHON_PATH = "/tmp/condaruntime/bin"
-        CONDA_PYTHON_RUNTIME = os.path.join(CONDA_PYTHON_PATH, "python")
+        runtime_meta = s3_client.head_object(Bucket=runtime_s3_bucket_used,
+                                             Key=runtime_s3_key_used)
+        ETag = str(runtime_meta['ETag'])[1:-1]
+        conda_python_path = "/tmp/condaruntime_{0}/bin".format(ETag)
```
shivaram: Use `CONDA_RUNTIME_DIR` ?

pywren/wrenhandler.py
```diff
@@ -25,11 +26,12 @@
...
     import version  # pylint: disable=relative-import
 
-PYTHON_MODULE_PATH = "/tmp/pymodules"
-CONDA_RUNTIME_DIR = "/tmp/condaruntime"
+PYTHON_MODULE_PATH = "/tmp/pymodules_{0}"
+CONDA_RUNTIME_DIR = "/tmp/condaruntime_{0}"
 RUNTIME_LOC = "/tmp/runtimes"
-JOBRUNNER_CONFIG_FILENAME = "/tmp/jobrunner.config.json"
-JOBRUNNER_STATS_FILENAME = "/tmp/jobrunner.stats.txt"
+JOBRUNNER_CONFIG_FILENAME = "/tmp/jobrunner_{0}.config.json"
```
shivaram: So this will get filled by `pid` right ? Can you include that in the comment ? It'll be good to know what is ETag specific and what is PID specific

tests/ec2_standalone_tests.sh
```diff
@@ -22,4 +22,11 @@ export PYWREN_EXECUTOR=remote
 pytest -v tests/test_simple.py
 RESULT=$?
 pywren standalone terminate_instances
+
+pywren standalone launch_instances 1 --max_idle_time=10 --idle_terminate_granularity=600 --pywren_git_commit=$TRAVIS_COMMIT  --parallelism 16
+sleep 20
+export PYWREN_EXECUTOR=remote
+pytest -v tests/test_simple.py
```
shivaram: Is there a test that actually checks if 16 were run in parallel ?

Vaishaal: Nope, I'm not really sure how to test that.

pywren/wrenhandler.py
```diff
@@ -52,42 +54,47 @@ def free_disk_space(dirname):
...
+    lock = open(RUNTIME_DOWNLOAD_LOCK, "a")
+    fcntl.lockf(lock, fcntl.LOCK_EX)
     # get runtime etag
     runtime_meta = s3_client.head_object(Bucket=runtime_s3_bucket,
                                          Key=runtime_s3_key)
     # etags have strings (double quotes) on each end, so we strip those
     ETag = str(runtime_meta['ETag'])[1:-1]
+    conda_runtime_dir = CONDA_RUNTIME_DIR.format(ETag)
     logger.debug("The etag is ={}".format(ETag))
     runtime_etag_dir = os.path.join(RUNTIME_LOC, ETag)
```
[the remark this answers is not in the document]
Vaishaal: yeah its because the runtime goes "inside" /tmp/runtimes/

pywren/wrenhandler.py
```diff
@@ -52,42 +54,47 @@ def free_disk_space(dirname):
...
+def download_runtime_if_necessary(s3_client, runtime_s3_bucket, runtime_s3_key,
+                                  delete_old_runtimes=False):
     """
     Download the runtime if necessary
 
     return True if cached, False if not (download occured)
 
     """
 
+    lock = open(RUNTIME_DOWNLOAD_LOCK, "a")
```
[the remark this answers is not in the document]
Vaishaal: no reason I think I'm just used to doing that for open calls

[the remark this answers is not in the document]
Vaishaal: you need (w, w+, or r)

Vaishaal:
@shivaram handled all your nits and added a test to make sure its actually running things in parallel

shivaram:
Thanks @Vaishaal - I'll take a look at the test later tonight. Lets also wait for @ericmjonas to take a look ?

Vaishaal:
ping

PR #226, opened 2017-11-22
Record container Id

ooq:
I'm redoing some measurement for container reuse.
This might also be useful to put into open source?
@ericmjonas @shivaram


shivaram:
I think @ericmjonas did a lot of testing on container reuse using other things like mac address / ntp offset etc. I'm just curious on how does this relate to those other tests ?

PR #227, opened 2017-11-28
Portability

### LINKED ISSUE #154: Refactor wrenhandler to be less lambda specific
This is related to [151](https://github.com/pywren/pywren/issues/151). 

Right now wrenhandler is hard coded to work specifically in the lambda environment. For cleanliness and for future portability, this could be refactored to be platform independent.

This includes:
Replacing existing calls to shell commands with python code.

Rewriting variables and commands to be portable between Windows and Linux environments (i.e. replacing `/tmp/pymodules` with `os.path.join(TEMP, "pymodules")`

Rewriting calls to s3 to use an abstract storage layer. I was thinking of using Qifan's storage class, but this would have to be extended to include calls to `head_object` and `get_object`.

Generally, for each cloud platform we've had to rewrite the wrenhandler in a different language, so I'm not sure what the scope of this should be.

apengwin:
This is to fix #154. The main changes here are

- Abstracting s3 IO to a storage handler.
- Removing any syscalls unavailable on windows.
- Making file paths portable across platforms.

This is an update of a previous PR (#155) which had these fixes, but there have been a lot of changes to the way pywren is structured, which made rebasing infeasible. This is essentially the same PR as earlier, except it follows the restructuring of `wrenhandler` and `jobrunner`.

### DIFF
--- a/pywren/jobrunner.py
+++ b/pywren/jobrunner.py
@@ -5,7 +5,7 @@ import shutil
 import json
 import sys
 import time
-import boto3
+from storage.storage import Storage
 
 
 from six.moves import cPickle as pickle
@@ -28,10 +28,9 @@ jobrunner_config_filename = sys.argv[1]
 
 jobrunner_config = json.load(open(jobrunner_config_filename, 'r'))
 
-
 # FIXME someday switch to storage handler
 # download the func data into memory
-s3_client = boto3.client("s3")
+storage_handler = Storage(jobrunner_config['storage_config'])
 
 func_bucket = jobrunner_config['func_bucket']
 func_key = jobrunner_config['func_key']
@@ -54,8 +53,8 @@ def write_stat(stat, val):
 
 try:
     func_download_time_t1 = time.time()
-    func_obj_stream = s3_client.get_object(Bucket=func_bucket, Key=func_key)
-    loaded_func_all = pickle.loads(func_obj_stream['Body'].read())
+    func_obj = storage_handler.get_object(func_key)
+    loaded_func_all = pickle.loads(func_obj)
     func_download_time_t2 = time.time()
     write_stat('func_download_time',
                func_download_time_t2-func_download_time_t1)
@@ -69,10 +68,17 @@ try:
 
     for m_filename, m_data in loaded_func_all['module_data'].items():
         m_path = os.path.dirname(m_filename)
+        if sys.platform == 'win32':
+            if len(m_path) > 0 and m_path[0] == "\\":
+                m_path = m_path[1:]
+            # fix windows forward slash delimeter
+            m_path = os.path.join(*filter(lambda x: len(x) > 0, m_path.split("/")))
 
-        if len(m_path) > 0 and m_path[0] == "/":
-            m_path = m_path[1:]
+        else:
+            if len(m_path) > 0 and m_path[0] == "/":
+                m_path = m_path[1:]
         to_make = os.path.join(PYTHON_MODULE_PATH, m_path)
+
         try:
             os.makedirs(to_make)
         except OSError as e:
@@ -93,16 +99,10 @@ try:
     # now unpickle function; it will expect modules to be there
     loaded_func = pickle.loads(loaded_func_all['func'])
 
-    extra_get_args = {}
-    if data_byte_range is not None:
-        range_str = 'bytes={}-{}'.format(*data_byte_range)
-        extra_get_args['Range'] = range_str
-
     data_download_time_t1 = time.time()
-    data_obj_stream = s3_client.get_object(Bucket=data_bucket,
-                                           Key=data_key, **extra_get_args)
+    data_obj = storage_handler.get_object(data_key, data_byte_range)
     # FIXME make this streaming
-    loaded_data = pickle.loads(data_obj_stream['Body'].read())
+    loaded_data = pickle.loads(data_obj)
     data_download_time_t2 = time.time()
     write_stat('data_download_time',
                data_download_time_t2-data_download_time_t1)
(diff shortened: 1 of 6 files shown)


pywren/jobrunner.py
```diff
@@ -5,7 +5,7 @@
 import json
 import sys
 import time
-import boto3
+from storage.storage import Storage # pylint: disable=relative-import
```
shivaram: Does this need to be relative or can we make it absolute ?

pywren/jobrunner.py
```diff
@@ -69,10 +68,17 @@ def write_stat(stat, val):
 
     for m_filename, m_data in loaded_func_all['module_data'].items():
         m_path = os.path.dirname(m_filename)
-
-        if len(m_path) > 0 and m_path[0] == "/":
-            m_path = m_path[1:]
+        if sys.platform == 'win32':
```
shivaram: Can we put this functionality into a util function ?

pywren/storage/s3_backend.py
```diff
@@ -24,7 +41,7 @@ def put_object(self, key, data):
         """
         self.s3client.put_object(Bucket=self.s3_bucket, Key=key, Body=data)
 
-    def get_object(self, key):
+    def get_object(self, key, data_byte_range=None):
```
shivaram: update the documentation below on what `data_byte_range` is and what its format should be ?

pywren/storage/storage.py
```diff
@@ -25,6 +25,19 @@ def __init__(self, config):
         else:
             raise NotImplementedError(("Using {} as storage backend is" +
                                        "not supported yet").format(config['storage_backend']))
+    def head_object(self, key):
+        """
+        Retrieves metadata for given key.
+        The metadata dict must have Contentlength and Etag as keys
```
shivaram: Is it `Contentlength` or `ContentLength` ? (or does it not matter)

pywren/wrenconfig.py
```diff
@@ -38,7 +38,9 @@ def load(config_filename):
                 config_filename, res['s3']['bucket']))
     if 'storage_backend' not in res:
         res = patch_storage_config(res)
-
+    elif res['storage_backend'] == 's3':
```
shivaram: could you remind me why we need this ? I remember @ericmjonas made the `patch_storage_config` for a specific reason. Also a comment here on what is going on would be useful ?

pywren/wrenhandler.py
```diff
@@ -139,18 +156,18 @@ def aws_lambda_handler(event, context):
...
     # containers in determining correct # of processors
     custom_handler_env = {'OMP_NUM_THREADS' : '1'}
+
     return generic_handler(event, context_dict, custom_handler_env)
 
 def get_server_info():
 
-    server_info = {'uname' : subprocess.check_output("uname -a", shell=True).decode("ascii")}
+    server_info = {'uname' : " ".join(platform.uname()),
+                   'cpuinfo': platform.processor()}
```
shivaram: do we use a space before `:` or not (the two lines are inconsistent). Also the `cpuinfo` is something we are adding new ?

pywren/wrenhandler.py
```diff
@@ -160,6 +177,8 @@ def generic_handler(event, context_dict, custom_handler_env=None):
     context_dict is generic infromation about the context
     that we are running in, provided by the scheduler
 
+    storage_handler is the storage object that we use for IO
```
shivaram: storage_handler is not a constructor argument ?

pywren/wrenhandler.py
```diff
@@ -237,20 +263,24 @@ def generic_handler(event, context_dict, custom_handler_env=None):
         response_status['call_id'] = call_id
         response_status['callset_id'] = callset_id
 
-        CONDA_PYTHON_PATH = "/tmp/condaruntime/bin"
+        if sys.platform == 'win32':
+            CONDA_PYTHON_PATH = r"D:\home\site\wwwroot\conda\Miniconda2"
```
shivaram: can we put this at the top of the file ? Also is this `D:` a windows standard temp path or is this just for Azure functions ?

pywren/wrenhandler.py
```diff
@@ -303,7 +341,10 @@ def consume_stdout(stdout, queue):
             if total_runtime > job_max_runtime:
                 logger.warning("Process exceeded maximum runtime of {} sec".format(job_max_runtime))
                 # Send the signal to all the process groups
-                os.killpg(os.getpgid(process.pid), signal.SIGTERM)
+                if sys.platform.startswith('linux'):
+                    os.killpg(os.getpgid(process.pid), signal.SIGTERM)
+                else:
+                    pass
```
shivaram: I'm guessing kill is not supported in windows ?

pywren/storage/storage.py
apengwin: I think it's `ContentLength`

pywren/wrenhandler.py
apengwin: `cpuinfo` replaces `/proc/cpuinfo`

pywren/wrenhandler.py
apengwin: This is just for azure functions.

pywren/wrenhandler.py
apengwin: I think this was because process group IDs work a lot different in windows. I'll look into this.

pywren/wrenconfig.py
apengwin: This is from #162. When we added the storage API, we added some field to`pywren_config` but didn't include them in the default config that's created when someone installs pywren.

pywren/jobrunner.py
apengwin: Did you mean line 75?

shivaram: I meant this whole block actually

PR #228, opened 2017-12-08
Move jobrunner to dir

ericmjonas:
When invoked, jobrunner doesn't depend on anything else in the pywren distribution. 

Due to the way python package lookups work, the current method to copy jobrunner into the runtime directory copies some extra packages and appends them to the path. Rather than manually manipulating the module path, it seemed like the right answer was to create a specific stand-alone jobrunner dir. All of the contents of that directory now are copied to the runner.


Vaishaal:
pretty simple change.

LGTM

PR #238, opened 2018-04-05
Add license headers.

ooq:
Add license headers to all `.py` files, except those under `./docs` and `./pywren/serialize/`

@ericmjonas @Vaishaal


shivaram:
Thanks @ooq - LGTM

ooq:
@shivaram I'm confused why it errs on the trailing lines. Looking into it.

shivaram:
@ooq why do we have trailing newlines ? can we not delete those newlines ?

ooq:
@shivaram The trailing newline error was due to a previously empty __init__ file. Fixed it. Also fixed error for files with special encoding. The push build is successful. I'll do a rebase to clean the commits.

PR #241, opened 2018-04-09
License update for serialize module

ooq:
Add the Multyvac BSD license header to the following files:
`pywren/serialize/default_preinstalls.py`
`pywren/serialize/module_dependency.py`

Keep cloudpickle BSD license header for `pywren/serialize/cloudpickle/cloudpickle.py`

Add PyWren BSD license header to all other files in `pywren/serialize/`

@shivaram

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,19 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+=======================================================================
+PyWren Subcomponents:
+
+The PyWren project contains subcomponents with separate copyright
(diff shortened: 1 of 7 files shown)


LICENSE
```diff
@@ -199,3 +199,19 @@
...
+=======================================================================
+PyWren Subcomponents:
+
+The PyWren project contains subcomponents with separate copyright
+notices and license terms. Your use of the source code for the these
+subcomponents is subject to the terms and conditions of the following
+licenses.
+
+========================================================================
+BSD-style licenses
```
shivaram: Can you explicitly note that some files are BSD 2 clause and some BSD 3 clause ? They are broadly similar but its better to be precise.

ooq: Good point! Done.

Issue #268, opened 2018-08-22
WaitTest::test_all_complete  fails in standalone mode some times due to botocore timeouts


Vaishaal:
Some times the wait test fails in standalone mode. I think this can be remedied by wrapping the boto3.read requests in an exp backoff loop. I can take care of it.

```==================================== FAILURES ===================================
__________________________ WaitTest.test_all_complete __________________________
self = <test_simple.WaitTest testMethod=test_all_complete>
    def test_all_complete(self):
        def wait_x_sec_and_plus_one(x):
            time.sleep(x)
            return x + 1
    
        N = 10
        x = np.arange(N)
    
        futures = pywren.default_executor().map(wait_x_sec_and_plus_one, x)
    
        fs_dones, fs_notdones = pywren.wait(futures,
                                        return_when=pywren.wren.ALL_COMPLETED)
>       res = np.array([f.result() for f in fs_dones])
tests/test_simple.py:276: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_simple.py:276: in <listcomp>
    res = np.array([f.result() for f in fs_dones])
../../../miniconda/envs/test-environment/lib/python3.5/site-packages/pywren-0.3.0-py3.5.egg/pywren/future.py:216: in result
    reraise(*self._traceback)
../../../miniconda/envs/test-environment/lib/python3.5/site-packages/six.py:692: in reraise
    raise value.with_traceback(tb)
/tmp/pywren.runner_4500/0/jobrunner.py:74: in <module>
    ???
/tmp/condaruntime_1f946a6b74d64a4b4ebd6bc928f033cf-15/lib/python3.5/site-packages/botocore/response.py:74: in read
    ???```

Vaishaal:
Fixed in #269

PR #269, opened 2018-08-22
add exp backoff in jobrunner get_object so it doesn't fail on start

Vaishaal:
The travis tests for #268  were non deterministic due to S3 read of function object failing due to a timeout in job runner. So I added a few lines of exponential backoff in jobrunner so this issue wouldn't happen.

### DIFF
--- a/pywren/jobrunner/jobrunner.py
+++ b/pywren/jobrunner/jobrunner.py
@@ -22,12 +22,15 @@ import json
 import sys
 import time
 import boto3
+from botocore.vendored.requests.packages.urllib3.exceptions import ReadTimeoutError
 
 
 from six.moves import cPickle as pickle
 from tblib import pickling_support
 
 pickling_support.install()
+BACKOFF = 1
+MAX_TRIES = 5
 
 
 def b64str_to_bytes(str_data):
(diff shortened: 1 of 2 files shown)


pywren/jobrunner/jobrunner.py
```diff
@@ -68,9 +71,23 @@ def write_stat(stat, val):
...
+            time.sleep(backoff)
+            backoff *= 2
+            num_tries += 1
+    return func_obj_stream
+
 try:
     func_download_time_t1 = time.time()
-    func_obj_stream = s3_client.get_object(Bucket=func_bucket, Key=func_key)
+
+    func_obj_stream = get_object_with_backoff(s3_client, bucket=func_bucket, key=func_key, max_tries=MAX_TRIES, backoff=BACKOFF, **extra_get_args)
```
shivaram: line is probably too long

pywren/jobrunner/jobrunner.py
```diff
@@ -68,9 +71,23 @@ def write_stat(stat, val):
...
+        except ReadTimeoutError:
+            time.sleep(backoff)
+            backoff *= 2
+            num_tries += 1
+    return func_obj_stream
+
 try:
     func_download_time_t1 = time.time()
-    func_obj_stream = s3_client.get_object(Bucket=func_bucket, Key=func_key)
+
```
shivaram: no need to pass in max_tries, backoff if its already the default argument ?

tests/test_simple.py
```diff
@@ -71,6 +71,8 @@ def throwexcept(x):
 
         with pytest.raises(Exception) as execinfo:
             res = fut.result() 
+
+        print("EXEC INFO ", str(execinfo.value))
```
shivaram: Is this for debugging ? Will you be deleting it before merging ?

pywren/jobrunner/jobrunner.py
```diff
@@ -115,8 +132,8 @@ def write_stat(stat, val):
         extra_get_args['Range'] = range_str
 
     data_download_time_t1 = time.time()
-    data_obj_stream = s3_client.get_object(Bucket=data_bucket,
-                                           Key=data_key, **extra_get_args)
+    data_obj_stream = get_object_with_backoff(s3_client, bucket=data_bucket,
```
shivaram: you need to pass the key here

Issue #270, opened 2018-08-30
Jobrunner not returning logfiles


rolloff:
My pywren code fails intermittently with either of the following two errors:
`  File "/home/ubuntu/pywren/pywren/future.py", line 139, in result
    raise self._exception
botocore.vendored.requests.packages.urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(67562629 bytes read, 12637 more expected)', IncompleteRead(67562629 bytes read, 12637 more expected))`

`  File "/home/ubuntu/pywren/pywren/future.py", line 139, in result
    raise self._exception
botocore.vendored.requests.packages.urllib3.exceptions.ReadTimeoutError: None: None`

I am trying to send a fairly large function to approximately ~1400 workers, so I thought that this might be caused by the JobRunner timing out when it tries to download the function from s3.  Vaishaal also confirmed that the size of the function is ~67MB.   

To verify that this is indeed the error, I added several logging.info and print statements inside the jobrunner, and exception handling around the s3 function download so that I would see a special message if the exception was raised. However, none of my messages are getting returned.  I don't see the messages when I run `pywren print_latest_logs` or when I look at the Cloudwatch web url. In addition, the messages don't show up when I call either ` print(finished_future.run_status["stdout"])` or `print(finished_future.invoke_status["stdout"])` from my local function.  

I know how to debug the large function if this is indeed the issue, but it would be very helpful to confirm that this is actually what's wrong, and for that I need the logging info.  Any help would be greatly appreciated! Thank you

andresriancho:
Having the same issue, the function's stdout does NOT show in cloudwatch.

andresriancho:
@rolloff any ideas on how to solve this issue?

Vaishaal:
Ahh this is a bug in  how we were consuming stdout in wrenhandler.

Fix is on the way.

https://github.com/pywren/pywren/commit/bac9bf3e1579e567f99b391b40ce00d5376d169d

andresriancho:
@Vaishaal the problem I'm seeing is that I see lines in the CloudWatch log written by [this call to logger.info](https://github.com/pywren/pywren/blob/master/pywren/wrenhandler.py#L391):

```python
logger.info("Running {} {}".format(time.time(), process.returncode))
```

But not [this other call to logger.info](https://github.com/pywren/pywren/blob/master/pywren/wrenhandler.py#L395) which is just a few lines below the first:

```python
                line = q.get_nowait()
                stdout += line
[...]

PR #277, opened 2018-09-09
fix wait on multiple callset_ids

### LINKED ISSUE #179: pywren.wait() only works for futures of same job
as title.

zehric:
Fix #179
Tested this fix very minimally, but it seems to work for both waiting on futures that are returned by different calls to map as well as different executors. Would like feedback on whether this is sufficient or if it could possibly create issues.


Vaishaal:
this LGTM.

zehric:
Travis builds are failing for a similar reason to #270

PR #282, opened 2018-09-16
clean up local invoker

zehric:
Made local invoker portable and semantically equivalent to a lambda invoker. This just involved making the local invoker's thread daemonic so that the program would exit when the main thread exits. ~However, it's possible there is a reason this was not done from the beginning, would like input on whether this is fine.~

Also added some tests, in test_local_invoker.py, which was just lifted from the dummy invoker tests.

### DIFF
--- a/pywren/invokers.py
+++ b/pywren/invokers.py
@@ -24,13 +24,17 @@ import shutil
 import tempfile
 import atexit
 import glob2
+import multiprocessing
+import atexit
 import botocore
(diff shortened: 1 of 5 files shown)


zehric:
Switched to multiprocessing to support true parallelism. Currently spawns a number of processes equal to the number of processors on the system as reported by `os.cpu_count()` and takes jobs from a synchronized queue, just as in the serial case.

PR #285, opened 2018-09-24
added sqs queue size and instance uptimes in cli

### LINKED ISSUE #215: Expose SQS queue length from command line
Since we have access to the SQS queue, we should be able to query the size of the queue (both "invisible" and  remaining) 

I was thinking something like:
```
$ pywren standalone queue_size
Jobs In Flight: 100
Jobs in Queue: 100

```

zehric:
resolves #215

Also decided to implement the `instance_uptime` command line option in the meantime

### DIFF
--- a/pywren/scripts/pywrencli.py
+++ b/pywren/scripts/pywrencli.py
@@ -561,9 +561,9 @@ def standalone_queue_size(ctx):
 
     sqs = boto3.resource('sqs', region_name=AWS_REGION)
     queue = sqs.get_queue_by_name(QueueName=SQS_QUEUE_NAME)
-    click.echo("Jobs in flight: {}".format(
+    click.echo("Approximate number of jobs in flight: {}".format(
         queue.attributes['ApproximateNumberOfMessagesNotVisible']))
-    click.echo("Jobs in queue: {}".format(
+    click.echo("Approximate number of jobs in queue: {}".format(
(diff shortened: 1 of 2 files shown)


pywren/scripts/pywrencli.py
```diff
@@ -541,6 +549,20 @@ def standalone_terminate_instances(ctx):
...
+@click.pass_context
+def standalone_queue_size(ctx):
+    config_filename = ctx.obj['config_filename']
+    config = pywren.wrenconfig.load(config_filename)
+    AWS_REGION = config['account']['aws_region']
+    SQS_QUEUE_NAME = config['standalone']['sqs_queue_name']
+
+    sqs = boto3.resource('sqs', region_name=AWS_REGION)
+    queue = sqs.get_queue_by_name(QueueName=SQS_QUEUE_NAME)
+    click.echo("Jobs in flight: {}".format(
```
Vaishaal: would probably add the word "Approximate" here

pywren/scripts/pywrencli.py
```diff
@@ -541,6 +549,20 @@ def standalone_terminate_instances(ctx):
...
+        queue.attributes['ApproximateNumberOfMessagesNotVisible']))
+    click.echo("Jobs in queue: {}".format(
```
Vaishaal: And here

Issue #284, opened 2018-09-24
standalone setup improvements


zehric:
I ran into a couple issues with the standalone setup when I first tried to use it.

- [x] on pywren-setup, if standalone is enabled, an error will be thrown if the default instance profile already exists (also noted in #254)
- [x] AMI is hardcoded in the default config and never changed. We should find a way to get a good AMI given the user's region
- [x] ssh key doesn't exist in default config, and a new key pair is never generated. ~#121 attempts to fix this but looks like it ran into some issues.~
- [x] launching spot instances currently requires the user to input a spot price manually. We should make a default option that makes the maximum price the on-demand price. 
    - we should (possibly) make spot pricing the default option for standalone mode if we do this
- [x] the cli command to launch an instance has a small bug; it passes in `max_idle_time` instead of `default_volume_size`

Issue #214, opened 2017-10-18
pywren standalone mode improvements


Vaishaal:
I would like to make standalone mode more useful. I can take point for most of these changes.

But what I'm thinking is: 
* Ability to automatically launch standalone instances from within python
* Ability to specify separate SQS queues for different pywren standalone executors 
* Instance termination from within python (via context manager?)
* Ability to monitor SQS queue load from command line
* Ability to have multiple executor on an ec2 machine

Vaishaal:
#215 covers SQS queue load issue

PR #286, opened 2018-09-30
Standalone executor

### LINKED ISSUE #217: Launch standalone cluster on pwex executor start
Basically a sane way of launching standalone machines from within python without breaking the serverless abstraction.

What @ericmjonas and I were thinking was something like:
```
pwex = pywren.standalone_executor(min_executors=32) # if there are < 32 executors, launch up to 32 executors, if there are > 32 executors, do nothing
```

zehric:
This resolves some of #214. Resolves #217.

The idea is that we make a subclass of `Executor` called `StandaloneExecutor`. The reason for this is that a standalone executor needs to keep more state than a normal executor, for example the instances that it launched as well as maybe the SQS queue that is associated with it. 

The standalone executor now takes two additional optional arguments, a `min_instances` and `spot_price`. `min_instances` works as in #217. `spot_price` is there because spot instances are usually much cheaper than normal instances and we should allow the user to launch spot instances from python as well. As stated in #284, I think that there should be a way to allow the user to go with the default spot price, but this isn't implemented yet.

I currently have a rough implementation of `StandaloneExecutor` with context managers. The context manager will terminate instances on exit, but there is also an exposed method `terminate_instances` which the user can call manually to terminate the instances associated with the given executor.


zehric:
Probably don't want to terminate instances on exit if a context manager isn't used.

zehric:
Now adds a way to set spot pricing to the demand price, by entering 0.

zehric:
Discussed changes:
- Remove purge queue on CTRL-C and instead make a new command for it
- Remove cancellation and cleanup and instead put it in `map` or `call_async`
- Remove auto-termination and context managers in general
- Add spot pricing to config and remove from `standalone_executor` params

PR #289, opened 2018-10-19
fix standalone setup errors

### LINKED ISSUE #254: create_instance_profile fails if profile already exists
This is similar to #236 but the error happens while setting up IAM for standalone mode. Stacktrace pasted below

```
Setting up standalone mode.
creating queue pywren-jobs-1
Traceback (most recent call last):
  File "/Users/shivaram/anaconda2/bin/pywren-setup", line 11, in <module>
    load_entry_point('pywren', 'console_scripts', 'pywren-setup')()
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/shivaram/debian-shared/pywren/pywren/scripts/setupscript.py", line 221, in interactive_setup
    ctx.invoke(pywrencli.create_instance_profile)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/shivaram/debian-shared/pywren/pywren/scripts/pywrencli.py", line 206, in create_instance_profile
    iam.create_instance_profile(InstanceProfileName=instance_profile_name)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/boto3/resources/factory.py", line 520, in do_action
    response = action(self, *args, **kwargs)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/boto3/resources/action.py", line 83, in __call__
    response = getattr(parent.meta.client, operation_name)(**params)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/botocore/client.py", line 251, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/shivaram/anaconda2/lib/python2.7/site-packages/botocore/client.py", line 537, in _make_api_call
    raise ClientError(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (EntityAlreadyExists) when calling the CreateInstanceProfile operation: Instance Profile pywren-standalone already exists.
```

### LINKED ISSUE #284: standalone setup improvements

zehric:
Resolves #284. Fixes #254.


pywren/scripts/pywrencli.py
```diff
@@ -209,10 +214,57 @@ def create_instance_profile(ctx):
     instance_profile_name = config['standalone']['instance_profile_name']
 
     iam = boto3.resource('iam')
-    iam.create_instance_profile(InstanceProfileName=instance_profile_name)
+    iamclient = boto3.client('iam')
+    response = iamclient.list_instance_profiles()
```
zehric: Travis doesn't have permissions to do this. Maybe we should use try/except after all?

shivaram: can we add IAM permissions for travis to do this ? @Vaishaal ?

zehric:
@shivaram I tested my changes manually. It works as long as you have instance profile permissions. If you don't, it will error on the very last part of setup. 

I also tested the automatic ec2 ssh key addition. Seems to work with all possible combinations of user inputs.

pywren/scripts/pywrencli.py
zehric: We migrated Travis to a new account and added the necessary permissions.

zehric:
Now resolves #284

Issue #267, opened 2018-08-22
Python 3.7 runtime


semitom:
Python 3.7 has been released on June 27th: https://www.python.org/downloads/release/python-370/

Pywren does not currently have a runtime for python 3.7:

```bash
> pywren create_config
No matching runtime package for python version  3.7
Python 2.7 runtime will be used for remote.
```

Vaishaal:
Ok we are on this.

Vaishaal:
We have added a 3.7 runtime to the default regions.

Vaishaal:
There are some major caveats with this runtime. 

* The minimum scipy version for py37 is 'scipy==1.1.0' (the old runtimes had 0.19.1) which unfortunately brings a lot of detritus (large shared objects). Even with a modified shrinkconda that aggressively removes sos. The runtime cannot be squeezed into the 512 MB disk space limitation. To remedy this the default runtime here *does not* include scipy. @shivaram @ericmjonas  any comments/ideas?

shivaram:
Could you provide a breakdown on how the size increases after shrinkconda has run with scipy 1.1.0 ? Is it close to 512MB or is it much bigger ? And if I'm right, we currently have no runtime for 3.7 right ? I'm trying to understand if this is a regression or a new but limited feature

Vaishaal:
No runtime for 3.7. So this a new but limited feature.

* Currently its 551 MB. 
Output of ``du -h --max-depth=3  | sort -h | tail -n30`` for python3.6 vs python3.7 in /tmp/condaruntime

3.7:
```
620K    ./share/info
660K    ./lib/python3.7/pydoc_data
836K    ./include/python3.7m
908K    ./lib/tk8.6/demos
912K    ./share/man/man1
1000K   ./lib/python3.7/lib2to3
1.1M    ./lib/sqlite3.21.0
1.6M    ./include/openssl
1.6M    ./lib/tcl8.6/encoding
1.7M    ./lib/python3.7/idlelib
1.8M    ./lib/python3.7/encodings
1.9M    ./lib/python3.7/ensurepip
1.9M    ./lib/python3.7/__pycache__
1.9M    ./lib/tk8.6
2.4M    ./compiler_compat
2.4M    ./lib/tcl8.6
3.3M    ./lib/python3.7/distutils
4.4M    ./share/doc/openssl
4.7M    ./share/doc
4.9M    ./include
5.0M    ./share/man/man3
5.3M    ./conda-meta
[...]

PR #293, opened 2018-11-03
Release0.4/new runtime buckets

Vaishaal:


Vaishaal:
Adding this for a clean paper-trail.

There were several problems with the old runtime buckets:

1. Old runtime buckets were owned by @ericmjonas and were of the form:
* pywren-public-us-$REGION
for the $REGION in : ['us-west-1', 'us-west-2', 'us-east-1', 'us-east-2', 'ap-northeast-2', 'eu-central-1',  'eu-west-1']
2.  However the runtime bucket for  'eu-central-1'  was 'pywren-public-eu-central-1-a and the bucket was completely empty. 
3.  And the runtime bucket 'eu-west-1' was *not* public and only had the minimal runtime. (So this was basically un-usable)
4. The ``aws s3 ls `` of some of these buckets showed a lot of broken runtimes.



This PR fixes this by 
 1. moving all the runtimes to
 pywren-runtimes-public-us-$REGION  (which is owned by a pywremaster account)
for the $REGION in : ['us-west-1', 'us-west-2', 'us-east-1', 'us-east-2', 'ap-northeast-2', 'eu-central-  'eu-west-1']
2. Removed list permissions (do we need this?)
3. We also added a python 3.7 runtime to all regions #267 but with some caveats

PR #294, opened 2018-11-05
Release0.4/rc0

Vaishaal:
Final changes for 0.4rc0
@shivaram @ericmjonas  can you review

### DIFF
--- a/pywren/version.py
+++ b/pywren/version.py
@@ -18,7 +18,7 @@ from __future__ import print_function
 # we're following the version number guidelines
 # from https://www.python.org/dev/peps/pep-0386/
 
-__version__ = "0.4.0"
+__version__ = "0.4.0rc0"
 
 if __name__ == "__main__":
     print(__version__)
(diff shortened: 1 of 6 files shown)


shivaram:
What were the travis problems with 3.7 ?

Vaishaal:
@shivaram fixed the 3.7 issues can you review please.

pywren/version.py
```diff
@@ -18,7 +18,7 @@
 # we're following the version number guidelines
 # from https://www.python.org/dev/peps/pep-0386/
 
-__version__ = "0.4.0"
+__version__ = "0.4.0rc0"
```
shivaram: should this be rc0 ?

Vaishaal: Not that I know of I'm not sure what our rc policy is but, since creating/moving tags in git is annoying, I'm going to create the 0.4 tag only when things are final.

shivaram: Just a warning that you shouldn't push to pypi without updating this ?

Issue #300, opened 2018-11-28
running `pywren deploy_lambda` fails due to missing jobrunner.py


nricklin:
```pywren deploy_lambda``` results in:

```
Traceback (most recent call last):
  File "/home/ubuntu/venv/bin/pywren", line 11, in <module>
    sys.exit(main())
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/pywren/scripts/pywrencli.py", line 711, in main
    return cli() # pylint: disable=no-value-for-parameter
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/ubuntu/venv/local/lib/python2.7/site-packages/pywren/scripts/pywrencli.py", line 299, in deploy_lambda
    zipfile_obj.write(f, arcname=a)
  File "/usr/lib/python2.7/zipfile.py", line 1123, in write
    st = os.stat(filename)
OSError: [Errno 2] No such file or directory: '/home/ubuntu/venv/local/lib/python2.7/site-packages/pywren/jobrunner/jobrunner.py'
```

jobrunner.py is in fact not in the package.

**Comment by nricklin** [comment]
Just a note that when I manually copy the jobrunner.py file into place from github it works, but I imagine you want to get it into the package properly.

```
git clone https://github.com/pywren/pywren
mkdir /home/ubuntu/venv/local/lib/python2.7/site-packages/pywren/jobrunner
cp pywren/pywren/jobrunner/jobrunner.py /home/ubuntu/venv/local/lib/python2.7/site-packages/pywren/jobrunner/jobrunner.py
```

[...]

PR #301, opened 2018-11-28
added __init__.py to jobrunner folder

Vaishaal:
When packaging python projects, one of the packaging pathways ignores all folders without a init.py. This causes the jobrunner.py to be missing in the release. This fixes issue #300

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -78,7 +78,7 @@ enable=
 # --disable=W"
 
 # These are errors still not addressed in PyWren.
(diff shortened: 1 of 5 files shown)


shivaram:
Just curious: What is the lint warning we ran into that needs a change of pylintrc ?

Vaishaal:
so this happened in travis:
```
pywren/serialize/cloudpickle/__init__.py:1:0: R0401: Cyclic import (pywren -> pywren.wren -> pywren.invokers -> pywren.local) (cyclic-import)
```

